PR #11, opened 2021-03-18
E2E testing scripts + Contract Adjustment

Remscar:
- E2E test scripts
- Deployment script for controller
- Helper script to add controller to the deployed registrar
- Minor code tweaks and fixes on the Registrar

### DIFF
--- a/contracts/Registrar.sol
+++ b/contracts/Registrar.sol
@@ -80,7 +80,7 @@ contract Registrar is
     address controller = msg.sender;
 
     // Domain parents must exist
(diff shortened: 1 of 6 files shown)


stan36:
The tests are really thorough. The only case I can think of that wasn't covered in the tests was in the "controllers" section(line 99 of registrar.spec.ts). Testing that multiple controllers will work might be useful for when we add the staking controller(although maybe this test would be better off added in the e2e.spec.ts).  I assume simply deploying two basic controllers would be enough to test this.

PR #12, opened 2021-03-19
ZNS Documentation

Remscar:
Took a first pass at ZNS documentation


README.md
```diff
@@ -1,8 +1,193 @@
...
+
+## Building and Development
+
+Follow this guide to learn how to get started with building and developing locally.
+
+This guide will assume you are using [Visual Studio Code](https://code.visualstudio.com/) as an IDE, however you can use whatever IDE you would like.
+
+### Required Tools
+
+This project requires [yarn](https://yarnpkg.com/) to be installed on your system, as well as [node.js](https://nodejs.org/en/download/).
```
benefacto, line 17: If we're having this sort of documentation, using [nvm](https://github.com/nvm-sh/nvm) would be very useful so we can document _which_ version of node.js we're expecting them to have. I can make an issue/card for this, if desired.

README.md
```diff
@@ -1,8 +1,193 @@
...
+> - Never use your personal wallet or mnemonic
+> - Always use a different mnemonic for each new mainnet deployment
+
+
+### Deploy Registrar
+
+The Registrar can must be deployed by running the command:
+
+```bash
+yarn hardhat run .\scripts\deploy-registrar.ts --network <network>
```
benefacto, line 120: Do these deployment and test scripts support deployment locally or an arbitrary RPC? If not, would it be difficult for them to do so? Once I have time to build a smart contract CI/CD (which I haven't lately), it may be useful to run them in the pipeline container itself to ensure that they work. I can make an issue/card for this, if desired.

benefacto (approved):
LGTM! I didn't know that _awhile_ was an English word, haha.

README.md
Remscar: I can configure a localhost network `local` or something

README.md
Remscar: I wanted to make it obvious that we use node and yarn, but not verbose to the point of specifying version. I'm thinking of leaving it as is for now

README.md
Remscar: You can now launch a local node using
`npx hardhat node`
And then connect to it using the `localhost` network

PR #18, opened 2021-04-06
Work/stan36/staking controller

stan36:
cleaned up logic for staking controller with gas efficiency in mind 

added logic for setting metadata, royalty amount and metadata locking

should be done unless additional functionality is decided upon


contracts/StakingController.sol
```diff
@@ -0,0 +1,142 @@
...
+  ContextUpgradeable,
+  ERC165Upgradeable,
+  ERC721HolderUpgradeable
+{
+
+  using ECDSAUpgradeable for bytes32;
+
+  IERC20Upgradeable private infinity;
+  IRegistrar private registrar;
+  uint256 private rootDomain;
```
Remscar: This isn't used anywhere, please remove it to save on storage gas costs.

contracts/StakingController.sol
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+    /**
+      @notice placeDomainBid allows a user to send a request for a new sub domain to a domains owner
+      @param requestHash is the hashed data for a domain request
+      @param ipfsHash is the IPFS hash containing the bids params(ex: name being requested, amount, stc)
+      @dev the IPFS hash must be emitted as a string here for the front end to be able to recover the bid info
+    **/
+    function placeDomainBid(
+      bytes32 requestHash,
```
Remscar: Is this the hash of the content at the ipfs location (dictated by the ipfsHash)
If so, wouldn't they be the same since the ipfs location is a hash of the content?

contracts/StakingController.sol
```diff
@@ -0,0 +1,142 @@
...
+      string memory ipfsHash
+    ) external {
+      emit DomainBidPlaced(
```
Remscar, line 82: How can this be tied back to the parent domain which the bid is being placed for to create a sub domain.

I could see the subgraph going and fetching the ipfs content, and then reading a `parentId` from there, but i'm unsure as to why `requestHash` is being emitted here.

contracts/StakingController.sol
```diff
@@ -0,0 +1,142 @@
...
+      );
+    }
+
+    /**
+      @notice approveDomainBid approves a domain bid, allowing the domain to be created.
+      @param parentId is the id number of the parent domain to the sub domain being requested
+      @param ipfsHash is the IPFS hash of the bids information
+      @param bidder is the address of the account that placed the bid being accepted
+    **/
+    function approveDomainBid(
```
Remscar, line 95: I was wrong in my discord message to you about the interfaces;

This function also needs to take the signed message from inside of the ipfs stored file (signature) and mark it as accepted

```
approvedBids[keccak256(abi.encode(ipfsHash))] = signature;
```

or perhaps

```
approvedBids[keccak256(abi.encode(ipfsHash))] = keccak256(signature);
```

(though i'm not sure how much the 2nd will actually save gas)

contracts/StakingController.sol
```diff
@@ -0,0 +1,142 @@
...
+  uint256 private rootDomain;
+
+  mapping(bytes32 => address) public approvedBids;
+
+  event DomainBidPlaced(
+    bytes32 indexed requestHash,
+    string indexed ipfsHash
+  );
+
+  event DomainBidAccepted(string indexed bidIdentifier);
```
Remscar: Rename to `DomainBidApproved` to match the function name

Remscar: This could be a `bytes32` to save gas

contracts/StakingController.sol
```diff
@@ -0,0 +1,142 @@
...
+    **/
+      function fulfillDomainBid(
+        uint256 parentId,
+        uint256 bidAmount,
+        uint256 royaltyAmount,
+        string memory metadata,
+        string memory name,
+        bytes memory signature,
+        bool lockOnCreation
+      ) external authorizedRecipient(metadata) {
```
Remscar: You are using the metadata as if it is the ipfs hash of the bid, which it is not. You're probably going to need to pass in another argument here `ipfsHash`

Remscar: Also, fulfilling a domain bid should be runnable by anyone, not just the bid placer. The bidder is the one who gets the domain, not the person fulfilling it.

contracts/StakingController.sol
```diff
@@ -0,0 +1,142 @@
...
+        address recoveredbidder = recover(keccak256(abi.encode(bidAmount, name, parentId, metadata)), signature);
+        address bidder = _msgSender();
+        address controller = address(this);
+        require(bidder == recoveredbidder, 'StakingController: incorrect bid info');
```
Remscar: The caller doesn't have to be the bidder (msgSender does not have to be the recoveredBidder)

contracts/StakingController.sol
```diff
@@ -0,0 +1,142 @@
...
+      function fulfillDomainBid(
+        uint256 parentId,
+        uint256 bidAmount,
+        uint256 royaltyAmount,
+        string memory metadata,
+        string memory name,
+        bytes memory signature,
+        bool lockOnCreation
+      ) external authorizedRecipient(metadata) {
+        address recoveredbidder = recover(keccak256(abi.encode(bidAmount, name, parentId, metadata)), signature);
```
Remscar: This signature cannot be trusted, it needs to be verified as correct from what is captured in the `approveBid` function.

Otherwise this function is vulnerable and it can currently be attacked:

1. User A makes a bid to create a subdomain of Domain X
2. Domain X owner approves the bid
3. User A passes the same `ipfsHash` (metadata in this function) as they placed the bid with, but sends a different signature + parameters than were in the ipfs stored file
4. User A is now able to create any domain, on any sub domain, at no price

contracts/StakingController.sol
```diff
@@ -0,0 +1,142 @@
...
+      @param parentId is the id number of the parent domain to the sub domain being requested
+      @param bidAmount is the uint value of the amount of infinity bid
+      @param royaltyAmount is the royalty amount the creator sets for resales on zAuction
+      @param metadata is the IPFS hash of the new domains information
+      @dev this is the same IPFS hash that contains the bids information as this is just stored on its own feild in the metadata
+      @param name is the name of the new domain being created
+      @param signature is the signature of the bidder
+      @param lockOnCreation is a bool representing whether or not the metadata for this domain is locked
+    **/
+      function fulfillDomainBid(
```
Remscar: Nothing is preventing this function from being called multiple times with the same bid since a bid is never 'destroyed' or 'expired'

contracts/StakingController.sol
```diff
@@ -0,0 +1,142 @@
...
+        }
+        emit DomainBidFulfilled(metadata);
+      }
+
+    /**
+      @notice recover allows the hashed data of a domain request to be recovered
+      @notice requestHash is the hash of the request being recovered
+      @notice signature is the signature the hash was created with
+    **/
+    function recover(
```
Remscar: After thinking about this more, this can probably just be made a private internal and removed from the interface since it's just a trivial method which other contracts would probably never call

contracts/StakingController.sol
```diff
@@ -0,0 +1,142 @@
...
+        string memory metadata,
+        string memory name,
+        bytes memory signature,
+        bool lockOnCreation
+      ) external authorizedRecipient(metadata) {
+        address recoveredbidder = recover(keccak256(abi.encode(bidAmount, name, parentId, metadata)), signature);
+        address bidder = _msgSender();
+        address controller = address(this);
+        require(bidder == recoveredbidder, 'StakingController: incorrect bid info');
+        infinity.transferFrom(bidder, controller, bidAmount);
```
Remscar: Using the SafeERC20 wrapper may be helpful here

Remscar (changes requested):
Looks like a great start.

Beyond the comments I left, I'd recommend avoiding naming things like `ipfsHash` and instead using a name that's more specific to what it is used for. For instance if it's the ipfs hash of the bid which is in ipfs, maybe `bidIpfsLocation` or `bidIpfsHash`.

Variables like `metadata` might also be more clear if it were `metadataUri`.

contracts/StakingController.sol
stan36: no, this is the signed message hash for the bid info

contracts/StakingController.sol
stan36: The request hash is being emitted here for easy retrieval by the parent ID owner so they can use recover to check that the bid info stored in IPFS matches the bid info of the signed message.

contracts/StakingController.sol
stan36: why? DomainBidAccepted is the name I got from the interface file you gave me.

contracts/StakingController.sol
stan36: Why would user A pay the gas for user B to claim their name? I have been trying to limit input params on this function as we will likely need to add additional ones later when we implement the dToken logic.

contracts/StakingController.sol
stan36: can you have multiples of the same domain name?

contracts/StakingController.sol
Remscar: Could we please rename this to something like `signedBid` then since `requestHash` is vague and doesn't accurately reflect what it is.

contracts/StakingController.sol
Remscar: Even by comparing those two signatures, you couldn't trust the bid. An attacker could put fake data in the bid JSON and then a signature for a different bid in the same JSON.

They could then upload the IPFS file, place the bid with the fake bid, but the signature doesn't match.
An event emitted with the signature here would match the one in the IPFS file, so the attack wouldn't be discovered.

Only if the domain owner were to verify the signature (using the recover method) in the IPFS file, with the bid parameters contained inside of it, would the attack be caught. Currently this check must be done off chain, we could build it in on a protocol level at the `approveBid` step though.

But because of that, emitting an event that has the signed message while also having the IPFS location is redundant since the IPFS location should also have the signed message.

contracts/StakingController.sol
Remscar: Also I believe that a signed message is longer than 32 bytes, so it's going to need to be a `string` type instead of `bytes32`

contracts/StakingController.sol
stan36: Not sure it is needed as we are only using infinity(and dTokens in the future) all of which are built using standard ERC20 contracts from openZeppelin if I remember correctly. Ill double check with Matt on this and change it if need be.

contracts/StakingController.sol
Remscar: You cannot, however this at best would be a bug, but at worse be an attack vector.

One attack for this could be that my bid is approved and then I fulfill it. Later, I release/delete the domain and get my staked tokens back. At any point, I could call this function again and re-fulfill my bid even though i've already done that once. Having a lifetime stake on a domain isn't a part of the design spec.

contracts/StakingController.sol
Remscar: There's a couple of reasons I could think of, one of which being an oracle system that automatically fulfills domain bids on behalf of users, but also as an opposite, why not?

We can remove this check and just use `recoveredBidder` instead of `bidder` and the function will operate exactly the same, except now anyone can fulfill domain bids.

contracts/StakingController.sol
Remscar: For consistency please, I must've forgotten to update that name when I sent that interface file to you.

contracts/StakingController.sol
Remscar: It's not a very big code change to add SafeERC20 and we've already been bit in the past by not using it. We could save ourselves, and possibly someone else some trouble in the future by using it here.

contracts/StakingController.sol
```diff
@@ -0,0 +1,146 @@
...
+      function fulfillDomainBid(
+        uint256 parentId,
+        uint256 bidAmount,
+        uint256 royaltyAmount,
+        string memory metadata,
+        string memory name,
+        bytes memory signature,
+        bool lockOnCreation
+      ) external {
+        bytes32 unsignedRequestHash  = approvedBids[keccak256(abi.encode(metadata))];
```
Remscar: Domain metadata is different than the bidIPFSHash which is what the key of this mapping is.

contracts/StakingController.sol
```diff
@@ -0,0 +1,146 @@
...
+      @param parentId is the id number of the parent domain to the sub domain being requested
+      @param bidIPFSHash is the IPFS hash of the bids information
+      @param unsignedRequestHash is the signed hashed data for a domain bid request
+    **/
+    function approveDomainBid(
+        uint256 parentId,
+        string memory bidIPFSHash,
+        bytes32 unsignedRequestHash
+    ) external authorizedOwner(parentId) {
+      approvedBids[keccak256(abi.encode(bidIPFSHash))] = unsignedRequestHash;
```
Remscar: The signature needs to be included here, either hashed and as the key, as the value of the mapping, or hashed and as the value of the mapping.

Without the signature, anyone can act on a bid.

contracts/StakingController.sol
```diff
@@ -0,0 +1,146 @@
...
+        bytes memory signature,
+        bool lockOnCreation
+      ) external {
+        bytes32 unsignedRequestHash  = approvedBids[keccak256(abi.encode(metadata))];
+        address recoveredbidder = recover(keccak256(abi.encode(bidAmount, name, parentId, metadata)), signature);
+        address controller = address(this);
+        require(unsignedRequestHash == keccak256(abi.encode(bidAmount, name, parentId, metadata)), 'StakingController: incorrect bid params');
+        infinity.transferFrom(recoveredbidder, controller, bidAmount);
+        address parentOwner = registrar.ownerOf(parentId);
+        uint256 id = registrar.registerDomain(parentId, name, controller, parentOwner);
```
Remscar: The creator of the domain is the `recoveredBidder` not the `parentOwner` for domains created through the staking controller.

contracts/StakingController.sol
```diff
@@ -0,0 +1,146 @@
...
+        uint256 parentId,
+        uint256 bidAmount,
+        uint256 royaltyAmount,
+        string memory metadata,
+        string memory name,
+        bytes memory signature,
+        bool lockOnCreation
+      ) external {
+        bytes32 unsignedRequestHash  = approvedBids[keccak256(abi.encode(metadata))];
+        address recoveredbidder = recover(keccak256(abi.encode(bidAmount, name, parentId, metadata)), signature);
```
Remscar: Please rename to `recoveredBidder`

contracts/StakingController.sol
Remscar: This is still attackable. It is possible to take someone else's bid and fulfill it with a signature that you signed instead, thus making the domain in your name instead of the bidders.

contracts/StakingController.sol
```diff
@@ -0,0 +1,146 @@
...
+        address controller = address(this);
+        require(unsignedRequestHash == keccak256(abi.encode(bidAmount, name, parentId, metadata)), 'StakingController: incorrect bid params');
```
Remscar: `keccak256(abi.encode(bidAmount, name, parentId, metadata))` should be refactored to a local variable

contracts/StakingController.sol
```diff
@@ -0,0 +1,146 @@
...
+        bytes32 unsignedRequestHash,
+        bytes memory signature
+      )
+      public
+      pure
+      returns (address) {
+        return unsignedRequestHash.toEthSignedMessageHash().recover(signature);
+      }
+
+
```
Remscar: Consider adding a public helper view method `createBid` which would take all the required parameters and return a hash of a bid, ie doing: `keccak256(abi.encode(bidAmount, name, parentId, metadata))`

package.json
```diff
@@ -51,5 +51,10 @@
   },
   "name": "zns-contracts",
   "version": "0.0.0",
-  "license": "MIT"
+  "license": "MIT",
+  "dependencies": {
+    "@eth-optimism/plugins": "^1.0.0-alpha.3",
+    "@nomiclabs/ethereumjs-vm": "^4",
+    "web3": "^1.2.9"
```
Remscar: Remove web3 dependency please

package.json
```diff
@@ -51,5 +51,10 @@
   },
   "name": "zns-contracts",
   "version": "0.0.0",
-  "license": "MIT"
+  "license": "MIT",
+  "dependencies": {
+    "@eth-optimism/plugins": "^1.0.0-alpha.3",
+    "@nomiclabs/ethereumjs-vm": "^4",
```
Remscar: Remove this dependency please

package.json
```diff
@@ -51,5 +51,10 @@
   },
   "name": "zns-contracts",
   "version": "0.0.0",
-  "license": "MIT"
+  "license": "MIT",
+  "dependencies": {
+    "@eth-optimism/plugins": "^1.0.0-alpha.3",
```
Remscar: Remove this dependency please

contracts/StakingController.sol
```diff
@@ -0,0 +1,146 @@
...
+import "@openzeppelin/contracts-upgradeable/proxy/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/utils/ContextUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/introspection/ERC165Upgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC721/ERC721HolderUpgradeable.sol";
+
+
+import "./interfaces/IRegistrar.sol";
+
+contract StakingController is
```
Remscar, line 15: Hashed and signed messages should implement EIP712 (https://eips.ethereum.org/EIPS/eip-712) for safety
https://medium.com/metamask/eip712-is-coming-what-to-expect-and-how-to-use-it-bb92fd1a7a26

An example of how this plays out can be seen in zAuction: https://github.com/zer0-os/zAuction/blob/master/contracts/zAuction.sol

contracts/mocks/Infinity.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.7.3;
+
+import "@openzeppelin/contracts-upgradeable/proxy/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
+
+
+contract Infinity is Initializable, ERC20Upgradeable {
```
Remscar: Rename to `MockToken` please

test/stakingcontroller.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+        bidSignature
+      );
+      expect(recoveredAddress).to.eq(user1.address);
+    });
+  });
+
+  describe("Places a bid", () => {
+    it("emits a DomainBidPlaced event with the correct bid info", async () => {
+
+      const bidRequestHash = await ethers.utils.keccak256(
```
Remscar: Consider refactoring this pattern into a helper

test/stakingcontroller.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+          bidRequestHash
+        );
+
+        expect(tx)
+          .to.emit(controller, "DomainBidApproved")
+          .withArgs(bidIPFSHash);
+        });
+      });
+
+      describe("FullFills a bid", () => {
```
Remscar: This needs failure test cases (scenarios where this will fail) such as:

- Fulfilling a bid that doesn't exist
- Fulfilling a bid with a wrong signature
- Another user signing the bid and trying to fulfill it
- Fulfilling a bid that has already been fulfilled
- Using a different 'metadata' IPFS location than the bid location

test/stakingcontroller.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+        bidIPFSHash
+      );
+
+      expect(tx)
+        .to.emit(controller, "DomainBidPlaced")
+        .withArgs(bidRequestHash, bidIPFSHash, bidSignature);
+      });
+    });
+
+    describe("Accepts a bid", () => {
```
Remscar: This needs failure cases tested such as:

- A user approving a bid to a domain they do not own

test/stakingcontroller.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      const recoveredAddress = await controllerAsUser1.recover(
+        bidRequestHash,
+        bidSignature
+      );
+      expect(recoveredAddress).to.eq(user1.address);
+    });
+  });
+
+  describe("Places a bid", () => {
+    it("emits a DomainBidPlaced event with the correct bid info", async () => {
```
Remscar, line 92: This needs failure test cases:

- A user placing a bid for a subdomain on a domain that doesn't exist

contracts/StakingController.sol
```diff
@@ -0,0 +1,146 @@
...
+    string bidIPFSHash,
+    bytes indexed signature
+  );
+
+  event DomainBidApproved(string indexed bidIdentifier);
+
+  event DomainBidFulfilled(string indexed bidIdentifier);
+
+  modifier authorizedOwner(uint256 domain) {
+    require(registrar.domainExists(domain), "Invalid Domain");
```
Remscar: Prefix these require messages with `Zer0 Naming Service:`

contracts/StakingController.sol
```diff
@@ -0,0 +1,146 @@
...
+        string memory metadata,
+        string memory name,
+        bytes memory signature,
+        bool lockOnCreation
+      ) external {
+        bytes32 unsignedRequestHash  = approvedBids[keccak256(abi.encode(metadata))];
+        address recoveredbidder = recover(keccak256(abi.encode(bidAmount, name, parentId, metadata)), signature);
+        address controller = address(this);
+        require(unsignedRequestHash == keccak256(abi.encode(bidAmount, name, parentId, metadata)), 'StakingController: incorrect bid params');
+        infinity.transferFrom(recoveredbidder, controller, bidAmount);
```
Remscar: Please use SafeERC20 to make the token transfer.

contracts/StakingController.sol
stan36: I feel this one is worth discussing a bit more as we are gonna hit a "stack too deep" error with the fulfillDomainBid function either by adding another string(and, if not) by adding another address field when we have to add in the dynamicToken logic. To overcome this I figured we could have the metadata for the domain being created contain the bid info for a bid making them the same IPFS hash. This bidInfo feild should be completely ignored by openSea so I dont see a problem with doing things this way.

contracts/StakingController.sol
```diff
@@ -0,0 +1,239 @@
...
+      @param bidIPFSHash is the IPFS hash that contains the bids information
+      @param name is the name of the sub domain being requested
+      **/
+      function createBid(
+        uint256 parentId,
+        uint256 bidAmount,
+        string memory bidIPFSHash,
+        string memory name
+      ) public pure returns(bytes32) {
+        return keccak256(abi.encode(bidAmount, name, parentId, bidIPFSHash));
```
Remscar: Nitpick: Order the arguments in `abi.encode` to be the same order they appear in the function arugments

contracts/StakingController.sol
Remscar: Use a struct for the input then, it will save on # of parameters

contracts/StakingController.sol
```diff
@@ -0,0 +1,239 @@
...
+      @dev this is the same IPFS hash that contains the bids information as this is just stored on its own feild in the metadata
+      @param name is the name of the new domain being created
+      @param signature is the signature of the bidder
+      @param lockOnCreation is a bool representing whether or not the metadata for this domain is locked
+    **/
+      function fulfillDomainBid(
+        uint256 parentId,
+        uint256 bidAmount,
+        uint256 royaltyAmount,
+        string memory metadata,
```
Remscar: I know there was a comment somewhere, but metadata can't be the same as the IPFS hash, but the solution to the `too many args` thing is to use a struct for some of these args

contracts/StakingController.sol
```diff
@@ -0,0 +1,239 @@
...
+      }
+
+      /**
+      @notice createBid is a pure function  that creates a bid hash for the end user
+      @param parentId is the ID of the domain where the sub domain is being requested
+      @param bidAmount is the amount being bid for the domain
+      @param bidIPFSHash is the IPFS hash that contains the bids information
+      @param name is the name of the sub domain being requested
+      **/
+      function createBid(
```
Remscar, line 200: This should be calling that internal `hash(Bid memory bid)` method you have. That's the hash we want to be using.

package.json
```diff
@@ -51,5 +51,8 @@
   },
   "name": "zns-contracts",
   "version": "0.0.0",
-  "license": "MIT"
+  "dependencies": {
```
Remscar: If these are needed, can you move these dependencies into `devDependencies` please

contracts/interfaces/IStakingController.sol
```diff
@@ -0,0 +1,81 @@
+pragma solidity ^0.7.3;
```
Remscar, line 2: Add `// SPDX-License-Identifier: MIT` to the top of this file please

contracts/StakingController.sol
```diff
@@ -0,0 +1,239 @@
+pragma solidity ^0.7.3;
```
Remscar, line 2: Add `// SPDX-License-Identifier: MIT` to the top of this file`

contracts/StakingController.sol
Remscar: Disregard please.

Remscar:
Almost there, just a few small comments and the removal of the EIP712 stuff should put it there for the most part. Just make sure to run the linter and fix any warnings/errors. You can use `yarn fix` to fix most of those automatically.

contracts/StakingController.sol
stan36: hmm, alright. I suppose I can salvage the Bid struct

package.json
stan36: I thought I got rid of these so not sure where its coming from. Ill try to hunt that down in the morning

contracts/interfaces/IStakingController.sol
stan36: I had it there but removed it because I figured thats where the license field in the json was coming from. I can re-add it but that may work against removing that field from the json.

package.json
stan36: yea, I just checked and its still removed in my version.

Remscar (approved):
lgtm

PR #19, opened 2021-04-21
Work/stan36/staking controller revamp

stan36:
finished contract rework so that all logic is on-chain  and tests(17 tests with failure cases tested as well as success).

### DIFF
--- a/contracts/StakingController.sol
+++ b/contracts/StakingController.sol
@@ -109,7 +109,8 @@ contract StakingController is
     function approveDomainBid(
         uint256 bidIdentifier
     ) external {
-      Bid memory bid = bids[bidIdentifier];
+      Bid storage bid = bids[bidIdentifier];
+      require(bid.bidAmount != 0, "ZNS: Bid doesnt exist");
       require(registrar.domainExists(bid.parentId), "ZNS: Invalid Domain");
       require(registrar.ownerOf(bid.parentId) == _msgSender(), "ZNS: Not Authorized Owner");
       bid.accepted = true;
@@ -131,13 +132,13 @@ contract StakingController is
       string memory metadata,
       bool lockOnCreation
     ) external {
-        Bid memory bid = bids[bidIdentifier];
-        require(bid.accepted == true, "ZNS: unnaccepted");
+        Bid storage bid = bids[bidIdentifier];
+        require(bid.accepted == true, "ZNS: bid not accepted");
         require(bid.approved == false, "ZNS: already fulfilled/withdrawn");
         infinity.safeTransferFrom(bid.bidder, controller, bid.bidAmount);
         uint256 id;
         bytes32 domainId;
-        if(!registrar.domainExists(bid.parentId)){
+        if(registrar.domainExists(bid.parentId)){
            id = registrar.registerDomain(bid.parentId, bid.name, controller, bid.bidder);
            domainId = keccak256(abi.encode(bid.parentId, bid.name));
            domainIdentifier[domainId] = id;
--- a/contracts/StakingController.sol
+++ b/contracts/StakingController.sol
@@ -138,7 +138,7 @@ contract StakingController is
         infinity.safeTransferFrom(bid.bidder, controller, bid.bidAmount);
         uint256 id;
         bytes32 domainId;
-        if(registrar.domainExists(bid.parentId)){
+        if(!registrar.domainExists(bid.parentId)){
            id = registrar.registerDomain(bid.parentId, bid.name, controller, bid.bidder);
            domainId = keccak256(abi.encode(bid.parentId, bid.name));
            domainIdentifier[domainId] = id;
--- a/contracts/StakingController.sol
+++ b/contracts/StakingController.sol
@@ -10,12 +10,15 @@ import "@openzeppelin/contracts-upgradeable/token/ERC721/ERC721HolderUpgradeable
 
 
 import "./interfaces/IRegistrar.sol";
+import "./interfaces/IStakingController.sol";
+
 
 contract StakingController is
   Initializable,
   ContextUpgradeable,
   ERC165Upgradeable,
-  ERC721HolderUpgradeable
+  ERC721HolderUpgradeable,
+  IStakingController
 {
 
   using SafeERC20Upgradeable for IERC20Upgradeable;
@@ -26,9 +29,6 @@ contract StakingController is
   uint256 public requestCount;
 
   mapping(uint256 => Request) public requests;
-  mapping(uint256 => bool) public tokensHeld;
-  mapping(uint256 => Request) public acceptedRequests;
-  mapping(bytes32 => uint256) public domainIdentifier;
 
   struct Request {
     uint256 parentId;
@@ -36,29 +36,8 @@ contract StakingController is
     address requester;
     string name;
     bool accepted;
-    bool fulfilled;
     bool valid;
   }
-
-  event DomainRequestPlaced(
-  uint256 indexed parentId,
-  uint256 indexed requestIdentifier,
-  uint256 requestAmount,
-  string indexed name,
-  address requestdder
-  );
-
-  event DomainRequestApproved(uint256 indexed requestIdentifier);
-
-  event DomainRequestFulfilled(
-    uint256 indexed requestIdentifier,
-    string name,
-    address recipient,
-    uint256 indexed domainId,
-    uint256 indexed parentID
-  );
-
-  event RequestWithdrawn(uint256 indexed requestIdentifier);
 
 
   function initialize(IRegistrar _registrar, IERC20Upgradeable _infinity) public initializer {
@@ -81,7 +60,7 @@ contract StakingController is
       uint256 _parentId,
       uint256 _requestAmount,
       string memory _name
-    ) external {
+    ) external override {
       require(registrar.domainExists(_parentId), "ZNS: Invalid Domain");
       requestCount++;
       requests[requestCount] = Request({
@@ -90,8 +69,7 @@ contract StakingController is
         requester: _msgSender(),
         name: _name,
         accepted: false,
-        fulfilled: false,
-        valid: false
+        valid: true
       });
 
       emit DomainRequestPlaced(
@@ -109,7 +87,7 @@ contract StakingController is
     **/
     function approveDomainRequest(
         uint256 requestIdentifier
-    ) external {
+    ) external override{
       Request storage request = requests[requestIdentifier];
       require(request.requestAmount != 0, "ZNS: Request doesnt exist");
       require(registrar.domainExists(request.parentId), "ZNS: Invalid Domain");
@@ -132,22 +110,19 @@ contract StakingController is
       uint256 royaltyAmount,
       string memory metadata,
       bool lockOnCreation
-    ) external {
+    ) external override{
         Request storage request = requests[requestIdentifier];
-        require(request.fulfilled == false, "ZNS: already fulfilled");
-        require(request.valid == false, "ZNS: request not valid");
+        require(request.valid == true, "ZNS: request not valid");
         require(request.accepted == true, "ZNS: request not accepted");
         uint256 domainId = registrar.registerDomain(request.parentId, request.name, controller, request.requester);
         registrar.setDomainMetadataUri(domainId, metadata);
         registrar.setDomainRoyaltyAmount(domainId, royaltyAmount);
         registrar.transferFrom(controller, request.requester, domainId);
-        acceptedRequests[domainId] = request;
         if (lockOnCreation) {
           registrar.lockDomainMetadataForOwner(domainId);
         }
-        request.fulfilled = true;
+        request.valid = false;
         infinity.safeTransferFrom(request.requester, controller, request.requestAmount);
-
         emit DomainRequestFulfilled(
           requestIdentifier,
           request.name,
@@ -163,11 +138,11 @@ contract StakingController is
       **/
       function withdrawRequest(
         uint256 requestIdentifier
-      ) external {
+      ) external override{
         Request storage request = requests[requestIdentifier];
         require(request.requester == _msgSender(), "ZNS: Not request creator");
         require(request.accepted == false, "ZNS: request already accepted");
-        request.valid = true;
+        request.valid = false;
         emit RequestWithdrawn(requestIdentifier);
       }
 
--- a/contracts/StakingController.sol
+++ b/contracts/StakingController.sol
@@ -133,8 +133,8 @@ contract StakingController is
       bool lockOnCreation
     ) external {
         Bid storage bid = bids[bidIdentifier];
-        require(bid.accepted == true, "ZNS: bid not accepted");
         require(bid.approved == false, "ZNS: already fulfilled/withdrawn");
+        require(bid.accepted == true, "ZNS: bid not accepted");
         infinity.safeTransferFrom(bid.bidder, controller, bid.bidAmount);
         uint256 id;
         bytes32 domainId;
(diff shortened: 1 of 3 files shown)


contracts/StakingController.sol
```diff
@@ -64,123 +73,132 @@ contract StakingController is
...
     ) external {
-      require(registrar.domainExists(parentId), "ZNS: Invalid Domain");
+      require(registrar.domainExists(_parentId), "ZNS: Invalid Domain");
+      bidCount++;
+      bids[bidCount] = Bid({
+        parentId: _parentId,
+        bidAmount: _bidAmount,
+        bidder: _msgSender(),
+        name: _name,
+        accepted: false,
```
Remscar, line 71: `accepted` and `approved` are confusing here since the actions are `place` `approve` `fulfill`

Maybe `accepted` can be changed to `fulfilled`?

contracts/StakingController.sol
```diff
@@ -64,123 +73,132 @@ contract StakingController is
...
+      string memory metadata,
+      bool lockOnCreation
+    ) external {
+        Bid storage bid = bids[bidIdentifier];
+        require(bid.approved == false, "ZNS: already fulfilled/withdrawn");
+        require(bid.accepted == true, "ZNS: bid not accepted");
+        infinity.safeTransferFrom(bid.bidder, controller, bid.bidAmount);
+        uint256 id;
+        bytes32 domainId;
+        if(!registrar.domainExists(bid.parentId)){
```
Remscar: I'm not sure what this code here is doing, or why there is a branch.

It seems that:

If the parent domain doesnt exist you are registering the child domain and then setting the 'domain identifier' of the child domain to the hash of the parent id and the child label.
Else if the parent domain does exist, you are getting the 'domain identifier' and then checking if tokens are being held.

When would this be called where the parent domain doesn't exist? And if the parent domain doesn't exist then this call should fail.

contracts/StakingController.sol
```diff
@@ -64,123 +73,132 @@ contract StakingController is
...
+           id = registrar.registerDomain(bid.parentId, bid.name, controller, bid.bidder);
+           domainId = keccak256(abi.encode(bid.parentId, bid.name));
```
Remscar: I think you're trying to get the domain Id of the child domain here, but this isn't how you do it.

Use this snippet:
```
uint256 labelHash = uint256(keccak256(bytes(name)));
uint256 domainId = uint256(keccak256(abi.encodePacked(parentId, labelHash)));
```

contracts/StakingController.sol
```diff
@@ -64,123 +73,132 @@ contract StakingController is
...
+        @notice withdrawBid allows a bidder to withdraw a placed bid should they change their mind
+        @param bidIdentifier is the number representing the bid being withdrawn
+      **/
+      function withdrawBid(
+        uint256 bidIdentifier
+      ) external {
+        Bid storage bid = bids[bidIdentifier];
+        require(bid.bidder == _msgSender(), "ZNS: Not bid creator");
+        require(bid.accepted == false, "ZNS: Bid already accepted");
+        bid.approved = true;
```
Remscar: This is deceptive, maybe create a different flag in the bid structure and call it `valid` and if the bid has been fulfilled or withdrawn, `valid` is set to false.

contracts/StakingController.sol
```diff
@@ -64,123 +73,132 @@ contract StakingController is
...
       **/
-      function createBid(
-        uint256 parentId,
-        uint256 bidAmount,
-        string memory bidIPFSHash,
-        string memory name
-      ) public pure returns(bytes32) {
-        return keccak256(abi.encode(parentId, bidAmount, bidIPFSHash, name));
+      function relenquishOwnership(
+        uint256 tokenId
```
Remscar: rename `tokenId` to `domainId`

contracts/StakingController.sol
```diff
@@ -64,123 +73,132 @@ contract StakingController is
...
+      ) external {
+        require(registrar.domainExists(tokenId), "ZNS: Invalid Domain");
+        require(registrar.ownerOf(tokenId) == _msgSender(), "ZNS: Not Authorized Owner");
+        registrar.transferFrom(_msgSender(), controller, tokenId);
```
Remscar: This isn't a safe way of relinquishing ownership because it just gives it back to the controller.

The domain NFT should be burned here, which means a call to the registrar, however there is no burn method, so that would need to be added.

contracts/StakingController.sol
```diff
@@ -64,123 +73,132 @@ contract StakingController is
...
+        Bid memory bid = acceptedBids[tokenId];
+        infinity.safeTransfer(bid.bidder, bid.bidAmount);
```
Remscar: `safeTransfer` needs to be done last here

contracts/StakingController.sol
```diff
@@ -64,123 +73,132 @@ contract StakingController is
...
+    function fulfillDomainBid(
+      uint256 bidIdentifier,
+      uint256 royaltyAmount,
+      string memory metadata,
+      bool lockOnCreation
+    ) external {
+        Bid storage bid = bids[bidIdentifier];
+        require(bid.approved == false, "ZNS: already fulfilled/withdrawn");
+        require(bid.accepted == true, "ZNS: bid not accepted");
+        infinity.safeTransferFrom(bid.bidder, controller, bid.bidAmount);
```
Remscar: `safeTransfer` needs to be done last here

contracts/StakingController.sol
```diff
@@ -64,123 +73,132 @@ contract StakingController is
...
+        uint256 tokenId
+      ) external {
+        require(registrar.domainExists(tokenId), "ZNS: Invalid Domain");
+        require(registrar.ownerOf(tokenId) == _msgSender(), "ZNS: Not Authorized Owner");
+        registrar.transferFrom(_msgSender(), controller, tokenId);
+        Bid memory bid = acceptedBids[tokenId];
+        infinity.safeTransfer(bid.bidder, bid.bidAmount);
+        tokensHeld[tokenId] = true;
+        if(registrar.isDomainMetadataLocked(tokenId)){
+          registrar.unlockDomainMetadata(tokenId);
```
Remscar: This call will always fail, if the domain metadata is locked, it will have been locked by the owner and the controller doesn't have authorization to unlock the metadata through this call.

Remscar (changes requested):
Left some preliminary comments. Please also change the concept of `bid` to `offer` or `request`.

There has been a great deal of confusion about `bidding` which implies this is an auction when it's more of a `request` or `offer` system.

contracts/StakingController.sol
stan36: yea, this was supposed to check if the domain name being created's ID exists already but I kinda masked the problem with the way I wrote the test, good catch

contracts/StakingController.sol
stan36: This actually helps a lot and means I can get rid of one of the mappings I made to fill a gap in knowledge here, thanks

contracts/StakingController.sol
stan36: Thats kinda why I did it this way. Why wouldnt it be safe to just allow the controller to hold a relinquished domain until someone else stakes it?

stan36: I guess I am just going to remove this functionality for now seeing as we cannot overcome the metadata locking problem easily

contracts/StakingController.sol
```diff
@@ -19,37 +18,47 @@ contract StakingController is
...
 
   IERC20Upgradeable private infinity;
   IRegistrar private registrar;
   address private controller;
+  uint256 public requestCount;
+
+  mapping(uint256 => Request) public requests;
+  mapping(uint256 => bool) public tokensHeld;
+  mapping(uint256 => Request) public acceptedRequests;
+  mapping(bytes32 => uint256) public domainIdentifier;
```
Remscar: Remove this please

contracts/StakingController.sol
```diff
@@ -19,37 +18,47 @@ contract StakingController is
...
-  using ECDSAUpgradeable for bytes32;
   using SafeERC20Upgradeable for IERC20Upgradeable;
 
   IERC20Upgradeable private infinity;
   IRegistrar private registrar;
   address private controller;
+  uint256 public requestCount;
+
+  mapping(uint256 => Request) public requests;
+  mapping(uint256 => bool) public tokensHeld;
```
Remscar: Remove this please (unused)

contracts/StakingController.sol
```diff
@@ -19,37 +18,47 @@ contract StakingController is
...
+  mapping(uint256 => Request) public acceptedRequests;
```
Remscar, line 30: Remove this please (set, but never read from)

contracts/StakingController.sol
```diff
@@ -63,123 +72,103 @@ contract StakingController is
...
-        registrar.transferFrom(controller, recoveredBidder, id);
+    function fulfillDomainRequest(
+      uint256 requestIdentifier,
+      uint256 royaltyAmount,
+      string memory metadata,
+      bool lockOnCreation
+    ) external {
+        Request storage request = requests[requestIdentifier];
+        require(request.fulfilled == false, "ZNS: already fulfilled");
+        require(request.valid == false, "ZNS: request not valid");
```
Remscar: Request should be valid (valid == true)

contracts/StakingController.sol
```diff
@@ -63,123 +72,103 @@ contract StakingController is
...
+      require(registrar.domainExists(_parentId), "ZNS: Invalid Domain");
+      requestCount++;
+      requests[requestCount] = Request({
+        parentId: _parentId,
+        requestAmount: _requestAmount,
+        requester: _msgSender(),
+        name: _name,
+        accepted: false,
+        fulfilled: false,
+        valid: false
```
Remscar: valid should be true to start

contracts/StakingController.sol
```diff
@@ -63,123 +72,103 @@ contract StakingController is
...
+      @param _requestAmount is the uint value of the amount of infinity request
+      @param _name is the name of the new domain being created
     **/
-    function placeDomainBid(
-      uint256 parentId,
-      bytes32 unsignedRequestHash,
-      bytes memory signature,
-      string memory bidIPFSHash
+    function placeDomainRequest(
+      uint256 _parentId,
```
Remscar: Please remove the `_` in the argument names so they are consistent with the rest of the file/codebase. Only use the `_` if it is needed.

contracts/StakingController.sol
```diff
@@ -63,123 +72,103 @@ contract StakingController is
...
+          registrar.lockDomainMetadataForOwner(domainId);
         }
-        approvedBids[hashOfSig] = false;
-        emit DomainBidFulfilled(
-          metadata,
-          name,
-          recoveredBidder,
-          id,
-          parentId
+        request.fulfilled = true;
```
Remscar: Also set `valid` to false

contracts/StakingController.sol
```diff
@@ -63,123 +72,103 @@ contract StakingController is
...
-        string memory bidIPFSHash,
-        string memory name
-      ) public pure returns(bytes32) {
-        return keccak256(abi.encode(parentId, bidAmount, bidIPFSHash, name));
+      function withdrawRequest(
+        uint256 requestIdentifier
+      ) external {
+        Request storage request = requests[requestIdentifier];
+        require(request.requester == _msgSender(), "ZNS: Not request creator");
+        require(request.accepted == false, "ZNS: request already accepted");
```
Remscar: This should instead be checking whether or not the request is valid, and only allow valid requests (`request.valid == true`)

contracts/StakingController.sol
```diff
@@ -63,123 +72,103 @@ contract StakingController is
...
+        request.valid = true;
```
Remscar: This should be setting valid to false

contracts/StakingController.sol
```diff
@@ -63,123 +72,103 @@ contract StakingController is
...
-        registrar.setDomainRoyaltyAmount(id, royaltyAmount);
-        registrar.transferFrom(controller, recoveredBidder, id);
+    function fulfillDomainRequest(
+      uint256 requestIdentifier,
+      uint256 royaltyAmount,
+      string memory metadata,
+      bool lockOnCreation
+    ) external {
+        Request storage request = requests[requestIdentifier];
+        require(request.fulfilled == false, "ZNS: already fulfilled");
```
Remscar: If you do my comment that is later in this file (set valid to false when fulfilled) then you can remove this

Remscar (changes requested):
Almost there

contracts/interfaces/IStakingController.sol
```diff
@@ -5,96 +5,64 @@ import "@openzeppelin/contracts-upgradeable/introspection/IERC165Upgradeable.sol
...
-    @param bidIPFSHash is the IPFS hash that contains the bids information
-    @param name is the name of the sub domain being requested
-    **/
-  function createBid(
-    uint256 parentId,
-    uint256 bidAmount,
-    string memory bidIPFSHash,
-    string memory name
-  ) external pure returns (bytes32);
+  function withdrawRequest(uint256 bidIdentifier) external;
```
Remscar, line 65: This should be removed

contracts/StakingController.sol
```diff
@@ -19,37 +18,47 @@ contract StakingController is
   ERC721HolderUpgradeable
```
Remscar: Make sure to have this implement `IStakingController` (use as a parent)

contracts/interfaces/IStakingController.sol
stan36: Why would I remove this? Every other function has an implementation in the interface so I don't understand why I would remove this one.

Remscar: Please disregard

contracts/StakingController.sol
```diff
@@ -11,44 +10,33 @@ import "@openzeppelin/contracts-upgradeable/token/ERC721/ERC721HolderUpgradeable
...
-    uint256 indexed parentID
-  );
-
-  modifier authorizedOwner(uint256 domain) {
-    require(registrar.domainExists(domain), "ZNS: Invalid Domain");
-    require(registrar.ownerOf(domain) == _msgSender(), "ZNS: Not Authorized Owner");
-    _;
+  struct Request {
+    uint256 parentId;
+    uint256 requestAmount;
```
Remscar: Consider renaming this to `offeredAmount` since it might be confusing that this is a request, but they are not requesting an amount, they are requesting a name

contracts/StakingController.sol
```diff
@@ -11,44 +10,33 @@ import "@openzeppelin/contracts-upgradeable/token/ERC721/ERC721HolderUpgradeable
...
+    address requester;
+    string name;
```
Remscar: Consider renaming to `requestedName` since they are requesting this name with this request

contracts/StakingController.sol
Remscar: Still not resolved

contracts/StakingController.sol
stan36: it wont let me resolve this for some dumb reason

contracts/StakingController.sol
```diff
@@ -63,123 +51,99 @@ contract StakingController is
...
-        bytes memory signature
-    ) external authorizedOwner(parentId) {
-      bytes32 hashOfSig = keccak256(abi.encode(signature));
-      approvedBids[hashOfSig] = true;
-      emit DomainBidApproved(bidIPFSHash);
+    function approveDomainRequest(
+        uint256 requestIdentifier
+    ) external override{
+      Request storage request = requests[requestIdentifier];
+      require(request.offeredAmount != 0, "ZNS: Request doesnt exist");
```
Remscar: A safer way to check if the request exists is to check `request.valid` instead. This is really just checking if the offered amount isn't zero, but maybe they want to only offer zero tokens

Remscar (approved):
lgtm

PR #32, opened 2021-11-04
Introduce set and lock function, change existing independent functions to internal

JamesEarle:


contracts/Registrar.sol
```diff
@@ -164,39 +164,18 @@ contract Registrar is
   }
 
   /**
-   * @notice Sets the domain metadata uri
-   * @param id The domain to set on
-   * @param uri The uri to set
-   */
-  function setDomainMetadataUri(uint256 id, string memory uri)
```
Remscar: This still needs to be an external function

contracts/Registrar.sol
```diff
@@ -164,39 +164,18 @@ contract Registrar is
...
-  }
-
-  /**
-   * @notice Locks a domains metadata uri
+   * @notice Both sets and locks domain metadata uri in a single call
    * @param id The domain to lock
+   * @param uri The uri to set
    * @param toLock whether the domain should be locked or not
    */
-  function lockDomainMetadata(uint256 id, bool toLock) external override {
```
Remscar, line 187: This still needs to be an external function

contracts/Registrar.sol
```diff
@@ -296,6 +275,38 @@ contract Registrar is
    * Internal Methods
    */
 
+  /**
+   * @notice Sets the domain metadata uri
+   * @param id The domain to set on
+   * @param uri The uri to set
+   */
+  function _setDomainMetadataUri(uint256 id, string memory uri) internal {
+    require(!isDomainMetadataLocked(id), "Zer0 Registrar: Metadata locked");
```
Remscar: `require` statements should generally be done by external functions and not done internally.

contracts/Registrar.sol
```diff
@@ -296,6 +275,38 @@ contract Registrar is
...
+    emit MetadataChanged(id, uri);
+  }
+
+  /**
+   * @notice Locks a domains metadata uri
+   * @param id The domain to lock
+   * @param toLock whether the domain should be locked or not
+   */
+  function _lockDomainMetadata(uint256 id, bool toLock) internal {
+    if (toLock) {
```
Remscar, line 308: These require statements should also be done in an external/public function instead of internally.

contracts/Registrar.sol
```diff
@@ -163,6 +163,24 @@ contract Registrar is
...
+  /**
+   * @notice Both sets and locks domain metadata uri in a single call
+   * @param id The domain to lock
+   * @param uri The uri to set
+   * @param toLock whether the domain should be locked or not
+   */
+  function setAndLockDomainMetadata(
+    uint256 id,
+    string memory uri,
+    bool toLock
```
Remscar: Remove `toLock` from this function. This is a combo of set + lock. Exposing `toLock` is confusing since it should always be true.

MattCharles (approved):
LGTM

contracts/Registrar.sol
```diff
@@ -296,6 +300,32 @@ contract Registrar is
...
+    } else {
+      require(isDomainMetadataLocked(id), "Zer0 Registrar: Not locked");
+      require(
+        domainMetadataLockedBy(id) == msg.sender,
+        "Zer0 Registrar: Not locker"
+      );
+    }
+  }
+
+  function _lockDomainMetadata(
```
Remscar: This function seems redundant, if it just calls `_setDomainLock` why don't we just call that instead of `_lockDomainMetadata`?

contracts/Registrar.sol
```diff
@@ -163,6 +163,22 @@ contract Registrar is
...
+   * @param id The domain to lock
+   * @param uri The uri to set
+   */
+  function setAndLockDomainMetadata(uint256 id, string memory uri)
+    external
+    override
+    onlyOwnerOf(id)
+  {
+    require(!isDomainMetadataLocked(id), "Zer0 Registrar: Metadata locked");
+    _validateLockDomainMetadata(id, true);
```
Remscar: Calling this is redundant;

calling `_validateLockDomainMetadata(id, true)` results in the following logic:
- require that `msg.sender` is the owner of domain `id`
- require that domain `id` is unlocked

However this logic has already happened:

- On line 174 you use the `onlyOwnerOf(id)` call modifier, which "requires that `msg.sender` is the owner of domain `id`"
- On line 176 you have the line `require(!isDomainMetadataLocked(id), "Zer0 Registrar: Metadata locked");` which "requires that domain `id` is unlocked" 

Thus you are repeating the two checks you've already performed.
Either remove the modifier and the require on line 176, or remove this function call to not perform the checks twice.

contracts/Registrar.sol
```diff
@@ -296,6 +300,32 @@ contract Registrar is
    * Internal Methods
    */
 
+  function _setDomainMetadataUri(uint256 id, string memory uri) internal {
```
Remscar, line 302: Excellent use of an internal method that abstracts the "set domain metadata uri" logic away, and is then is used by two external methods to perform this logic (DRY principle ✔️ )

JamesEarle: thanks!

Remscar (approved):
🚢 :shipit:

PR #36, opened 2022-01-05
Added bulk mint function for one user

MattCharles:


RyanCross (approved):

contracts/BasicController.sol
```diff
@@ -17,8 +17,8 @@ contract BasicController is
   ERC721HolderUpgradeable
 {
   IRegistrar private registrar;
-  uint256 rootDomain; // for upgrade reasons
-  address admin; // not writeable - can mint a domain at any level in zNS
+  uint256 public rootDomain; // for upgrade reasons
+  address private admin; // not writeable - can mint a domain at any level in zNS
```
Remscar: If you make this `private` it means you can't read it from Etherscan / other contracts (easily) (There still is a way to do it, so it's not actually hidden)

`public` variables can be read externally, but not written to externally. `public` variables can be written to if another smart contract *inherits* this one. So if you made a `BasicController2` that extends/inherits this `BasicController`, the BasicController2 contract could write to the `admin` variable defined inside of the `BasicController`

contracts/BasicController.sol
```diff
@@ -94,6 +94,49 @@ contract BasicController is
     }
   }
 
+  function mintDomainsBulk(
```
Remscar: Can you make this a new controller? We still need the plain old generic Basic Controller, this added functionality deserves being it's own controller.

contracts/BasicController.sol
Remscar: Nit: This "admin" / "owner" construct has already been done before. You could take advantage of Open Zeppelins `Ownable` contracts here. https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/access/OwnableUpgradeable.sol

All you would need to do is import `OwnableUpgradeable`, inherit from it, and then call `__Ownable_init()` in the initialize method on this contract, example of this can be found in `zns-sales`: https://github.com/zer0-os/zNS-sales/blob/d420e2e7da1c663fc7befbf770818a1a63e8ab9b/contracts/SimpleSale.sol#L24

PR #39, opened 2022-01-13
Authorized Basic Controller

Remscar:
Added a Basic Controller which only allows authorized users to mint domains.
For now it allows global minting abilities.
Also added a new function on the Registrar which will mint + send a domain to a user. This is a common use case.

### DIFF
--- a/contracts/AuthBasicController.sol
+++ b/contracts/AuthBasicController.sol
@@ -10,7 +10,7 @@ import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 import "./interfaces/IBasicController.sol";
 import "./interfaces/IRegistrar.sol";
 
-contract BasicController is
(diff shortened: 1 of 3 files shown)


contracts/AuthBasicController.sol
```diff
@@ -0,0 +1,77 @@
...
+    }
+    authorizedAccounts[account] = isAuthorized;
+    emit AccountAuthorizationChanged(account, isAuthorized);
+  }
+
+  function registerSubdomainExtended(
+    uint256 parentId,
+    string memory label,
+    address owner,
+    string memory metadata,
```
RyanCross: Nit: Since this might be interacted with via etherscan, could this also be `metadataUri` so a user knows the expected format of the metadata?

RyanCross (approved):
One super nit but otherwise 👍

PR #43, opened 2022-02-22
Added function to set domain metadata in bulk using an ordered list of IDs and a folder.

MattCharles:


Remscar (approved):

contracts/Registrar.sol
```diff
@@ -414,4 +414,69 @@ contract Registrar is
...
+    uint256 royaltyAmount,
+    bool locked
+  ) external onlyController returns (uint256[] memory) {
+    uint256[] memory results;
+    for (uint256 i = 0; i < totalToRegister; i++) {
+      results[i] = _registerDomain(
+        parentId,
+        uint2str(i + namingOffset),
+        minter,
+        string(abi.encodePacked(folderWithIPFSPrefix, uint2str(i))),
```
JamesEarle, line 446: `uint2str` below looks like what's in [this stackoverflow post](https://stackoverflow.com/questions/47129173/how-to-convert-uint-to-string-in-solidity) which tells me there are possible issues with that code. A better and cleaner alternative is using the below answer on that post (or right here)

```
import "@openzeppelin/contracts/utils/Strings.sol";

Strings.toString(myUint)
```

JamesEarle (commented):
Would probably switch to using an OZ contract for uint -> string conversion, for security and simplicity

contracts/Registrar.sol
MattCharles: So in order to import that contract, I'd have to upgrade our contracts to at least `solidity 0.8.0`, which then conflicts with `OwnableUpgradeable.sol`, which has to be strictly less than `0.8.0`. How would you feel about doing a run on a testnet, making sure the int->str conversions work, and then putting an item on the backlog to bring contracts up to `0.8.0`? I would really rather not get bogged down in upgrading our contract versions seeing as we're trying to mint the remaining wheels today.

MattCharles: here's the `toString` function as defined in that Strings contract - https://github.com/OpenZeppelin/openzeppelin-contracts/blob/52eeebecda140ebaf4ec8752ed119d8288287fac/contracts/utils/Strings.sol#L15-L35 I could also just pull this implementation if we want

Remscar:
Might close this since 1.1 changes are gonna be big and we can just merge this into that

PR #45, opened 2022-04-14
Wolf Launch

Remscar:


contracts/registrar/ZNSHub.sol
```diff
@@ -106,6 +107,7 @@ contract ZNSHub is
     require(!authorizedRegistrars[registrar], "ZH: Already Registered");
 
     authorizedRegistrars[registrar] = true;
+    subdomainRegistrars[rootDomainId] = registrar;
```
MattCharles, line 110: are we gonna have to go back and re-add a bunch of registrars after this?

MattCharles (approved):

contracts/registrar/ZNSHub.sol
Remscar: None were there yet so it's okay

PR #46, opened 2022-04-19
Folder groups

Remscar:


contracts/registrar/Registrar.sol
```diff
@@ -25,6 +25,14 @@ contract Registrar is
     uint256 royaltyAmount;
     uint256 parentId;
     address subdomainContract;
+    // This is the folder group the domain belongs to
```
RyanCross, line 28: I'm assuming this is for IPFS but it would be nice to give some context to the concept of a folder group either here or in the readme somewhere.

contracts/registrar/ZNSHub.sol
```diff
@@ -32,6 +32,20 @@ contract ZNSHub is
...
+    address registrar,
+    uint256 indexed id,
+    string label,
+    uint256 indexed labelHash,
+    uint256 indexed parent,
+    address minter,
+    address controller,
+    string metadataUri,
+    uint256 royaltyAmount,
+    uint256 groupId,
```
RyanCross, line 45: nit: `folderGroupId` & `folderGroupIndex`?

contracts/registrar/ZNSHub.sol
```diff
@@ -53,6 +67,12 @@ contract ZNSHub is
     address childRegistrar
   );
 
+  event EEDomainGroupUpdatedV1(
```
RyanCross, line 70: This domain group concept, is it the idea that every registrar will have its own folderGroup? Will the fileGroupIndexes on a domain ever need to change when a group is updated?

contracts/registrar/ZNSHub.sol
```diff
@@ -53,6 +67,12 @@ contract ZNSHub is
...
+    address parentRegistrar,
+    uint256 folderGroupId,
+    string baseUri
```
RyanCross: base`What`Uri?

scripts/v1.1/upgrade-hub.ts
```diff
@@ -2,7 +2,9 @@ import * as hre from "hardhat";
 import { Registrar__factory, ZNSHub__factory } from "../../typechain";
 
 // mainnet Hub
-const hubAddress = "0x3F0d0a0051D1E600B3f6B35a07ae7A64eD1A10Ca";
+//const hubAddress = "0x3F0d0a0051D1E600B3f6B35a07ae7A64eD1A10Ca";
+// rinkeby hub
+const hubAddress = "0x90098737eB7C3e73854daF1Da20dFf90d521929a";
```
RyanCross, line 7: Should update this value in https://github.com/zer0-os/zer0-contracts and then consume the package.

contracts/registrar/ZNSHub.sol
Remscar: We had a lengthy discussion about this weeks ago and decided to go with just `groupId` to keep it generic

contracts/registrar/ZNSHub.sol
Remscar: Not every registrar has it's own group. Sets of domains in a registrar may be in a group together. Multiple groups can exist on a registrar.

Remscar: The indexes of domains in a group will never change.

contracts/registrar/ZNSHub.sol
Remscar: for the metadata.

scripts/v1.1/upgrade-hub.ts
Remscar: In a different PR we can do this

contracts/registrar/Registrar.sol
Remscar: Will add a good documentation pass to the backlog

RyanCross (commented):
Looks good. I feel like I have an understanding of the changes needed to make the necessary updates for the data store.

I am worried that non passive changes in smart contracts (new events, the addition of a new event hub address) could easily slip by and start causing problems for the data store.  Would you be open to adding a PR template in smart contract repos that directs the user to open an issue in the event-listener repo whenever a change like this is made?

EDIT: Nvm on the contract address I was mistaken. It would be good to get a better comms system for new event types though

RyanCross (approved):

PR #50, opened 2022-04-21
Folder group tests

JamesEarle:
Very minor contract changes as well

### DIFF
--- a/contracts/registrar/Registrar.sol
+++ b/contracts/registrar/Registrar.sol
@@ -81,7 +81,7 @@ contract Registrar is
    * @param baseUri The entire base uri (include ipfs://.../)
    */
   function updateDomainGroup(uint256 id, string memory baseUri)
-    private
+    external
     onlyController
   {
     require(id != 0 && id <= numDomainGroups, "Folder group invalid");
@@ -327,8 +327,11 @@ contract Registrar is
     uint256 labelHash = uint256(keccak256(bytes(label)));
 
     // Calculate the new domain's id and create it
+    // We must include a domain's group ID to avoid
+    // this hash creating conflicts with other domains
+    // at the same index in a different group.
     uint256 domainId = uint256(
-      keccak256(abi.encodePacked(parentId, labelHash))
+      keccak256(abi.encodePacked(parentId, labelHash, groupId))
     );
 
     // Create not inside of a domain group
--- a/contracts/registrar/Registrar.sol
+++ b/contracts/registrar/Registrar.sol
@@ -327,11 +327,8 @@ contract Registrar is
     uint256 labelHash = uint256(keccak256(bytes(label)));
 
     // Calculate the new domain's id and create it
-    // We must include a domain's group ID to avoid
-    // this hash creating conflicts with other domains
-    // at the same index in a different group.
     uint256 domainId = uint256(
-      keccak256(abi.encodePacked(parentId, labelHash, groupId))
+      keccak256(abi.encodePacked(parentId, labelHash))
     );
 
     // Create not inside of a domain group
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -48,9 +48,13 @@ const config: HardhatUserConfig = {
     hardhat: {
       accounts: [
         {
-          privateKey: `0x${process.env.MAINNET_PRIVATE_KEY}`,
+          privateKey: `0x${process.env.ASTRO_PRIVATE_KEY}`,
           balance: "10000000000000000000000",
         },
+        {
+          privateKey: `0x${process.env.MAIN_PRIVATE_KEY}`,
+          balance: "10000000000000000000000"
+        }
       ],
       forking: {
         url: "https://mainnet.infura.io/v3/fa959ead3761429bafa6995a4b25397e",
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -46,16 +46,6 @@ const config: HardhatUserConfig = {
   },
   networks: {
     hardhat: {
-      accounts: [
-        {
-          privateKey: `0x${process.env.ASTRO_PRIVATE_KEY}`,
-          balance: "10000000000000000000000",
-        },
-        {
-          privateKey: `0x${process.env.MAIN_PRIVATE_KEY}`,
-          balance: "10000000000000000000000"
-        }
-      ],
       forking: {
         url: "https://mainnet.infura.io/v3/fa959ead3761429bafa6995a4b25397e",
       },
--- a/test/folder-groups.spec.ts
+++ b/test/folder-groups.spec.ts
@@ -1,15 +1,18 @@
 import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/dist/src/signer-with-address";
 import { ethers, upgrades } from "hardhat";
+import chai from "chai";
+import { solidity } from "ethereum-waffle";
+import { BigNumber } from "ethers";
+import * as smock from "@defi-wonderland/smock";
+
 import {
   Registrar,
   ZNSHub,
   ZNSHub__factory,
   Registrar__factory,
 } from "../typechain";
-import chai from "chai";
-import { solidity } from "ethereum-waffle";
-import { BigNumber } from "ethers";
-import * as smock from "@defi-wonderland/smock";
+
+import { domainNameToId } from "./helpers";
 
 chai.use(solidity);
 const { expect } = chai;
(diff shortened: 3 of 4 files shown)


contracts/registrar/Registrar.sol
```diff
@@ -327,8 +327,11 @@ contract Registrar is
     uint256 labelHash = uint256(keccak256(bytes(label)));
 
     // Calculate the new domain's id and create it
+    // We must include a domain's group ID to avoid
+    // this hash creating conflicts with other domains
+    // at the same index in a different group.
     uint256 domainId = uint256(
-      keccak256(abi.encodePacked(parentId, labelHash))
+      keccak256(abi.encodePacked(parentId, labelHash, groupId))
```
Remscar: How is this going to effect existing domains and their ID's that have been created?
What is the formula I use to calculate a domain's id now?

When will this conflict happen? I'm not sure why this change needs to be made or how this prevents conflicts with other domains at the same index in a different group. Where does the domains index in a group come into this, I only see the Parent Domain Id, and the hash of the domain label being hashed, nothing to do with the group index.

test/folder-groups.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    const numDomainGroups = await registry.numDomainGroups();
+    expect(numDomainGroups.eq("0"))
+  });
+  it("creates a domain group", async () => {
+    await registry.addController(controller.address);
+
+    const asController = registry.connect(controller);
+
+    // Test folders
+    const uri1 = "ipfs://QmafuzfZ2doheWYL9tDLm2t39vZvtjfnPy1QyHX4HVawqN";
```
Remscar: This is an incorrect uri format to use for a domain group.

test/folder-groups.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    const uri2 = "ipfs://QmRrp9Hichv1jV1SxkSx8pqnhKhjHMpPZFKmtKi8C8pALQ";
+
+    let retrievedUri1 = await asController.domainGroups("1");
+    expect(retrievedUri1 === uri);
+
+    const retrievedUri2 = await asController.domainGroups("2");
+    expect(retrievedUri2 === uri2);
+
+    // Was private but never referred to by anything? make public
+    await asController.updateDomainGroup(ethers.BigNumber.from("1"), updatedUri);
```
Remscar: You can just pass `1` or `"1"` here instead of `ethers.BigNumber.from`.

test/folder-groups.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    retrievedUri1 = await asController.domainGroups("1");
+    expect(retrievedUri1 === updatedUri);
+  });
+  it("registers domains in a domain group", async () => {
+    const asController = registry.connect(controller);
+    const controllerAddress = await controller.getAddress()
+
+    let params = {
+      parentId: ethers.BigNumber.from("0"),
```
Remscar: You don't need to use `ethers.BigNumber.from` you can just use `0` or `"0"` to make this more readable.

Remscar: If you see something that says `BigNumberish` it doesn't mean you need to give it a big number, just something similar to a big number.

test/folder-groups.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      params.locked
+    );
+
+    const receipt = await tx.wait();
+    console.log(receipt.logs.length);
+
+    // Mint in second group
+    params.groupId = ethers.BigNumber.from("2");
+
+    const tx2 = await asController.registerDomainInGroupBulk(
```
Remscar: You are registering a duplicated set of domains here.

In the first call you register domains

`0://0, 0://1, 0://2, 0://3, 0://4, 0://5`

and then in the second call you once again try to register the domains:

`0://0, 0://1, 0://2, 0://3, 0://4, 0://5`

But this time with a different domain group id.

Because you modified the code in `Registrar.sol` it no longer will throw an error. But you've allowed for the same domain name to be registered twice.

Remscar: The basic registrar tests in `registrar.spec.ts` don't seem to work anymore. (Have to remove the accounts from the 'hardhat' network in `hardhat.config.ts` to make the tests run though)
image

test/folder-groups.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+  });
+  it("Gets a domain from a domain group", async () => {
+    const updatedUri = "ipfs://QmTfoSpX2JjLJrccCkAQ6Doh4Pwd47HxwKfSXmhW5j9ojM";
+
+    const domainUri = await registry.domainGroups("1")
+    expect(domainUri === updatedUri);
+
+    // Root and then 5 each for 2 groups = 11 total
+    const supply = await (await registry.totalSupply()).toNumber();
+    console.log(supply);
```
Remscar: Should we have an `expect` instead of a console log?

test/folder-groups.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    for (let i = 0; i < supply; i++) {
+      const domainId = await registry.tokenByIndex(i);
+      const domain = await registry.records(domainId);
+      console.log(
```
Remscar: Can we be doing an `expect` here?

test/folder-groups.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+        `domainId: ${domainId.toString()}, domainGroup: ${domain.domainGroup.toString()}, groupIndex ${domain.domainGroupFileIndex.toString()}`
+      );
+    }
+  });
+  it("Transfers a domain", async () => {
```
Remscar: This is already tested in other test files

hardhat.config.ts
```diff
@@ -48,9 +48,13 @@ const config: HardhatUserConfig = {
     hardhat: {
       accounts: [
         {
-          privateKey: `0x${process.env.MAINNET_PRIVATE_KEY}`,
```
Remscar, line 51: Would be best to not include these changes. If you're trying to get tests running just remove the `accounts` field of this and hardhat will give you 10 accounts by default.

test/folder-groups.spec.ts
JamesEarle: if modifying the `groupId` we pass into `registerDomainInGroupBulk` is not enough to make it mint domains within that group, what else is required? I got retrieved all the domains in the following test and that output looked right

Remscar: What the the names of the domains you've minted though? Read my above comment. Are you minting domains with the same name twice?

test/folder-groups.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    for (let i = 0; i < supply; i++) {
+      const domainId = await registry.tokenByIndex(i);
+      const domain = await registry.records(domainId);
+      console.log(
+        `domainId: ${domainId.toString()}
+        \nparentId: ${domain.parentId.toString()}
+        \ndomainGroup: ${domain.domainGroup.toString()}
+        \ngroupIndex ${domain.domainGroupFileIndex.toString()}\n`
+      );
```
Remscar: Could we add a test that does the following scenario;

- Update domain group 1's uri
- Checks a domain in that group to make sure the `tokenUri` has changed properly

test/folder-groups.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    const domainUri = await registry.domainGroups("1")
+    expect(domainUri === updatedUri);
+
+    // Root, 3 in group 1, and 2 in group 2
+    const supply = await (await registry.totalSupply()).toNumber();
+    expect(supply === 6)
+
+    for (let i = 0; i < supply; i++) {
+      const domainId = await registry.tokenByIndex(i);
```
Remscar: `tokenByIndex` does not return to you domains like 0://1 0://2
It's going to return to you domains based on the order they were minted.

ie.
`tokenByIndex(0)` will return you 0://
`tokenByIndex(1)` will return you 0://0
`tokenByIndex(2)` will return you 0://1
`tokenByIndex(3)` will return you 0://2

If you want to get the domain id of a domain based on it's name you want to call `domainNameToId()` which is in the `helpers/` (test/helpers)

`domainNameToId("0")` will return the id of 0://0
`domainNameToId("1")` will return the id of 0://1
`domainNameToId("wilder.wheels")` will return the id of 0://wilder.wheels
`domainNameToId("wilder.wheels.0")` will return the id of 0://wilder.wheels.0

`tokenByIndex` is typically dangerous to use unless you have a specific reason to get tokens by the order they were created in the registrar.

PR #52, opened 2022-05-12
Matt/wolf deployment

Remscar:


Remscar (commented):
Remove/archive `colorOutput.txt` and `newColoredIDs.json`

PR #53, opened 2022-05-17
Domain Purchasing (network)

Remscar:
Added the `DomainPurchaser.sol` smart contract which allows for users to mint (purchase) a subdomain of another domain.

Currently it only works for purchasing network domains (eg. 0://meow, 0://kitty)

### DIFF
--- a/contracts/controllers/DomainPurchaser.sol
+++ b/contracts/controllers/DomainPurchaser.sol
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.11;
+
+import {OwnableUpgradeable, Initializable} from "../oz/access/OwnableUpgradeable.sol";
+import {IRegistrar} from "../interfaces/IRegistrar.sol";
+import {IZNSHub} from "../interfaces/IZNSHub.sol";
+import {IERC20Upgradeable} from "../oz/token/ERC20/IERC20Upgradeable.sol";
+import {SafeERC20Upgradeable} from "../oz/token/ERC20/SafeERC20Upgradeable.sol";
+
+contract DomainPurchaser is Initializable, OwnableUpgradeable {
+  using SafeERC20Upgradeable for IERC20Upgradeable;
+
+  struct PricingData {
+    uint256 long;
+    uint256 medium;
+    uint256 short;
+  }
+
+  event NetworkPurchased(uint256 networkDomainId, address owner);
+
+  IERC20Upgradeable public paymentToken;
+  IZNSHub public zNSHub;
+  address public proceedsWallet;
+  PricingData public prices;
+
+  function initialize(
+    address _paymentToken,
+    address _zNSHub,
+    address _proceedsWallet,
+    PricingData memory _prices
+  ) public initializer {
+    __Ownable_init();
+    paymentToken = IERC20Upgradeable(_paymentToken);
+    zNSHub = IZNSHub(_zNSHub);
+    proceedsWallet = _proceedsWallet;
+    prices = _prices;
+  }
+
+  function purchaseNetwork(string memory name, string memory metadataUri)
+    external
+  {
+    // Purchased networks will always be on the root (0://) domain
+    uint256 rootDomainId = 0;
+
+    uint256 lengthOfName = strlen(name);
+    require(lengthOfName > 0, "DP: Empty string");
+    require(lengthOfName < 33, "DP: Name too long");
+
+    // Calculate out the cost of the network domain
+    uint256 price = prices.long; // Default price
+    if (lengthOfName < 4) {
+      price = prices.short;
+    } else if (lengthOfName < 8) {
+      price = prices.medium;
+    }
+
+    // Take ZERO tokens (ERC20) from the user
+    paymentToken.safeTransferFrom(msg.sender, proceedsWallet, price);
+
+    // Create the network domain
+    IRegistrar rootRegistrar = zNSHub.getRegistrarForDomain(rootDomainId);
+    uint256 createdDomainId = rootRegistrar.registerSubdomainContract(
+      rootDomainId,
+      name,
+      msg.sender, // Minter
+      metadataUri,
+      0, // Royalty Amount
+      false, // is Metadata Locked
+      msg.sender // Who to send the domain
+    );
+
+    emit NetworkPurchased(createdDomainId, msg.sender);
+  }
+
+  /**
+   * Calculates out the length of a string
+   */
+  function strlen(string memory s) internal pure returns (uint256) {
+    uint256 len = 0;
+    uint256 bytelength = bytes(s).length;
+    for (uint256 i = 0; i < bytelength; len++) {
+      bytes1 b = bytes(s)[i];
+      if (b < 0x80) {
+        i += 1;
+      } else if (b < 0xE0) {
+        i += 2;
+      } else if (b < 0xF0) {
+        i += 3;
+      } else if (b < 0xF8) {
+        i += 4;
+      } else if (b < 0xFC) {
+        i += 5;
+      } else {
+        i += 6;
+      }
+    }
+    return len;
+  }
+}
--- a/contracts/controllers/DomainPurchaser.sol
+++ b/contracts/controllers/DomainPurchaser.sol
@@ -4,63 +4,172 @@ pragma solidity ^0.8.11;
 import {OwnableUpgradeable, Initializable} from "../oz/access/OwnableUpgradeable.sol";
 import {IRegistrar} from "../interfaces/IRegistrar.sol";
 import {IZNSHub} from "../interfaces/IZNSHub.sol";
+import {IDomainPurchaser} from "../interfaces/IDomainPurchaser.sol";
 import {IERC20Upgradeable} from "../oz/token/ERC20/IERC20Upgradeable.sol";
 import {SafeERC20Upgradeable} from "../oz/token/ERC20/SafeERC20Upgradeable.sol";
 
-contract DomainPurchaser is Initializable, OwnableUpgradeable {
+contract DomainPurchaser is
+  IDomainPurchaser,
+  Initializable,
+  OwnableUpgradeable
+{
   using SafeERC20Upgradeable for IERC20Upgradeable;
 
-  struct PricingData {
-    uint256 long;
-    uint256 medium;
-    uint256 short;
+  // Names that are < shortNameLength are considered 'short'
+  uint256 constant shortNameLength = 4;
+  // Names that are < mediumNameLength are 'medium'
+  uint256 constant mediumNameLength = 8;
+  // Names that are < maxNameLength are 'long'
+  uint256 constant maxNameLength = 33;
+
+  // Divison basis for 'platformFee' percentages
+  uint256 constant divisonBasisPlatformFee = 1000;
+
+  /**
+   * Configuration / State of purchasing related information for a domain
+   */
+  struct DomainPurchaseData {
+    /**
+     * Whether or not subdomain minting is enabled on this domain.
+     * If this is set, `.prices` needs to be also set.
+     */
+    bool subdomainMintingEnabled;
+    /**
+     * Pricing information for minting subdomains
+     */
+    PricingData prices;
+    /**
+     * Whether or not subdomains minted on this domain are allowed to mint subdomains
+     * Example:
+     *    if 'allowSubdomainsToMint' is false, I (the owner of 0://cats) could allow another user to mint
+     *    '0://cats.meow' but then that user could not have others mint subdomains on that such as:
+     *    '0://cats.meow.tuna'
+     *    if 'allowSubdomainsToMint' is true, then '0://cats.meow.tuna' could be minted if the owner of
+     *    '0://cats.meow' enables subdomain minting on ''0://cats.meow' (ie 'subdomainMintingEnabled')
+     *
+     * Note: If a subdomain is minted while 'allowSubdomainsToMint' is enabled, then it inherits that
+     *    state and ignores the parent domains 'allowSubdomainsToMint' state.
+     */
+    bool allowSubdomainsToMint;
+    /**
+     * Copy of `.allowSubdomainsToMint` when this domain was minted.
+     * This way if a user mints a subdomain on a domain which did allow subdomain minting
+     * but then later that (parent) domain disables it
+     * The user is still able to allow others to mint subdomains on that domain
+     */
+    bool wasAllowedToSubdomainMintOnCreation;
   }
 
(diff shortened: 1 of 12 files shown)


contracts/controllers/DomainPurchaser.sol
```diff
@@ -0,0 +1,335 @@
...
+        i += 3;
+      } else if (b < 0xF8) {
+        i += 4;
+      } else if (b < 0xFC) {
+        i += 5;
+      } else {
+        i += 6;
+      }
+    }
+    return len;
```
JamesEarle, line 354: Add a more detailed comment above this function with some explanation

contracts/controllers/DomainPurchaser.sol
```diff
@@ -0,0 +1,335 @@
...
+
+  function initialize(
+    address _paymentToken,
+    address _zNSHub,
+    address _platformWallet,
+    PricingData memory _rootPrices,
+    uint256 _platformFee
+  ) public initializer {
+    __Ownable_init();
+    paymentToken = IERC20Upgradeable(_paymentToken);
```
JamesEarle, line 88: if it is possible for a network to change their payment token in zAuction shouldn't we also allow that here? Is there a way to keep them in sync? If they are intentionally not to be changed in this contract, why?

contracts/controllers/DomainPurchaser.sol
```diff
@@ -0,0 +1,335 @@
...
+    if (price > 0) {
+      if (parentId == 0) {
+        // The platform is always considered to own the root domain (parentId == 0)
+        paymentToken.safeTransferFrom(msg.sender, platformWallet, price);
+      } else {
+        uint256 fee = (price * platformFee) / divisonBasisPlatformFee;
+
+        // Take ZERO tokens (ERC20) from the user
+        paymentToken.safeTransferFrom(
+          msg.sender,
```
JamesEarle, line 164: if we know `price > 0` there should be a `require(paymentToken.balanceOf(msg.sender) >= price, "...")` before we do a transfer

contracts/controllers/DomainPurchaser.sol
```diff
@@ -0,0 +1,335 @@
...
+      "DP: Not authorized"
+    );
+    DomainPurchaseData storage record = purchaseData[domainId];
+
+    bool differentPrices = record.prices.short != pricing.short ||
+      record.prices.medium != pricing.medium ||
+      record.prices.long != pricing.long;
+    require(
+      differentPrices,
+      "Pricing info is the same. Use .setDomainMintingStatus"
```
JamesEarle: nit: start error message with `"DP:..."` for consistency

contracts/controllers/DomainPurchaser.sol
```diff
@@ -0,0 +1,335 @@
...
+  ) external returns (uint256) {
+    DomainPurchaseData memory domainData = purchaseData[parentId];
+
+    // This will be removed later when the feature becomes available
+    if (parentId != 0) {
+      revert("DP: Only purchasing network domains is allowed currently.");
+    }
+
+    require(domainData.subdomainMintingEnabled, "DP: Minting not enabled");
+
```
JamesEarle, line 123: if `subdomainMintingEnabled` is true, as stated above we also need to ensure prices are set, but the below seems to go fine in the case that they are not? If free domain minting is fine is it temporary? or should the comment above (or the code below) be changed?

test/domainPurchaser.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    it("allows a user to purchase a network", async () => {
+      hub.getRegistrarForDomain.returns(registrar.address);
+      registrar.registerSubdomainContract.returns(1);
+
+      erc20Token.transferFrom.returns(true);
+
+      const tx = await purchaser
+        .connect(user1)
+        .purchaseSubdomain(0, "meowmeowmeow", "ipfs://Qm1");
```
JamesEarle, line 74: image

test/domainPurchaser.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+      registrar.registerSubdomainContract.returns(1);
+
+      erc20Token.transferFrom.reset();
+      erc20Token.transferFrom.reverts();
+
+      const tx = purchaser
+        .connect(user1)
+        .purchaseSubdomain(0, "👻👻👻", "ipfs://Qm1", { gasLimit: 3000000 });
+
+      await expect(tx).to.be.revertedWith("SafeERC20: low-level call failed");
```
JamesEarle, line 102: Left a comment above, shouldn't we just check the user balance ourselves?

contracts/controllers/DomainPurchaser.sol
```diff
@@ -0,0 +1,335 @@
...
+    if (parentId != 0) {
```
Will-Hyde: If this whole block of code is unreachable, is it worth including it? Or maybe make the revert command higher up toggleable?

contracts/controllers/DomainPurchaser.sol
```diff
@@ -0,0 +1,335 @@
...
+        // The platform is always considered to own the root domain (parentId == 0)
+        paymentToken.safeTransferFrom(msg.sender, platformWallet, price);
+      } else {
+        uint256 fee = (price * platformFee) / divisonBasisPlatformFee;
+
+        // Take ZERO tokens (ERC20) from the user
+        paymentToken.safeTransferFrom(
+          msg.sender,
+          zNSHub.ownerOf(parentId),
+          price - fee
```
Will-Hyde, line 166: If the fee is at 100% it will be transfering 0,
probably not an issue but I know you made sure to check for 0 in other spots of the code before transferring

JamesEarle (commented):
Some questions if you could clarify please

Will-Hyde (approved):

contracts/controllers/DomainPurchaser.sol
Remscar: It is not possible for networks to change their payment token in current spec. Only ZERO is usable.

contracts/controllers/DomainPurchaser.sol
Remscar: `safeTransferFrom` reverts if the user doesn't have enough tokens, so checking balance beforehand results in wasted gas.

test/domainPurchaser.spec.ts
Remscar: No, we don't want to add the extra gas cost.

contracts/controllers/DomainPurchaser.sol
Remscar: The above revert statement will be removed in a later version making this code reachable. When it is removed, this is the right place for it to be. I might make the above statement toggleable though.

contracts/controllers/DomainPurchaser.sol
Remscar: I cant imagine the fee every being 100% so I might make it max 99.9%

contracts/controllers/DomainPurchaser.sol
Remscar: Free domain minting may be intentional, so pricing does not need to be set.

JamesEarle: Okay, make sure to update the comment that says otherwise on line 34 of this file then please.

contracts/interfaces/IDomainPurchaser.sol
```diff
@@ -0,0 +1,58 @@
...
+  ) external returns (uint256);
+
+  /**
+   * Sets the purchase price of a domain, and allows for enabling subdomain minting to be set
+   * @param domainId The domain to set pricing data for
+   * @param pricing The pricing data
+   * @param subdomainMintingEnabled Whether subdomain minting should be enabled
+   * @param allowSubdomainsToMint Whether to allow subdomains to mint their own subdomains via the purchaser
+   */
+  function setDomainPricing(
```
RyanCross, line 39: The job of setting the purchase price/pricing data of a domain and the job of setting subdomain mint settings seem like two very different responsibilities. 

Is this wrapped into a single function due to smart contract limitations or to save on gas?

contracts/controllers/DomainPurchaser.sol
```diff
@@ -0,0 +1,357 @@
...
+  uint256 public platformFee;
+
+  // Mapping of Domain Id => DomainPurchaseData
+  mapping(uint256 => DomainPurchaseData) public purchaseData;
+
+  // Allow for non-network domains to be minted
+  bool allowMintingNonNetwork = false;
+
+  function initialize(
+    address _paymentToken,
```
RyanCross, line 81: What's the intent behind the underscore naming convention?

contracts/controllers/DomainPurchaser.sol
```diff
@@ -0,0 +1,357 @@
...
+  // Divison basis for 'platformFee' percentages
+  uint256 constant divisonBasisPlatformFee = 1000;
+
+  /**
+   * Configuration / State of purchasing related information for a domain
+   */
+  struct DomainPurchaseData {
+    /**
+     * Whether or not subdomain minting is enabled on this domain.
+     * If this is set, `.prices` needs to be also set.
```
RyanCross: super nit: remove the period in `.prices`

contracts/controllers/DomainPurchaser.sol
```diff
@@ -0,0 +1,357 @@
...
+     *    if 'allowSubdomainsToMint' is false, I (the owner of 0://cats) could allow another user to mint
+     *    '0://cats.meow' but then that user could not have others mint subdomains on that such as:
+     *    '0://cats.meow.tuna'
+     *    if 'allowSubdomainsToMint' is true, then '0://cats.meow.tuna' could be minted if the owner of
+     *    '0://cats.meow' enables subdomain minting on ''0://cats.meow' (ie 'subdomainMintingEnabled')
+     *
+     * Note: If a subdomain is minted while 'allowSubdomainsToMint' is enabled, then it inherits that
+     *    state and ignores the parent domains 'allowSubdomainsToMint' state.
+     */
+    bool allowSubdomainsToMint;
```
RyanCross, line 52: Prior to us locking down subdomain minting, any subdomain owner could mint a new domain, right? I believe under the hood this called the registrar controllers `registerSubdomainExtended` method.

How will this behavior affect/integrate with that method and the workflow of minting subdomains?

test/domainPurchaser.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+    it("charges the user the right amount for short length network", async () => {
+      hub.getRegistrarForDomain.returns(registrar.address);
+      registrar.registerSubdomainContract.returns(1);
+
+      erc20Token.transferFrom.reset();
+      erc20Token.transferFrom.returns(true);
+
+      await purchaser
+        .connect(user1)
+        .purchaseSubdomain(0, "👻👻👻", "ipfs://Qm1");
```
RyanCross, line 114: Setting the subdomain name as spooky ghosties got me thinking. Do we need logic to disallow certain characters? At the very least including the domain delimiter character `.` should throw an error

contracts/controllers/DomainPurchaser.sol
```diff
@@ -0,0 +1,357 @@
...
+
+  // Names that are < shortNameLength are considered 'short'
+  uint256 constant shortNameLength = 4;
+  // Names that are < mediumNameLength are 'medium'
+  uint256 constant mediumNameLength = 8;
+  // Names that are < maxNameLength are 'long'
+  uint256 constant maxNameLength = 33;
+
+  // Divison basis for 'platformFee' percentages
+  uint256 constant divisonBasisPlatformFee = 1000;
```
RyanCross: *Division

Also, for a mathematically challenged fellow like myself, could you explain the purpose of this division basis?  Previously I understood platform fees as just a percentage of the asking price.

contracts/interfaces/IDomainPurchaser.sol
Remscar: Yes

contracts/controllers/DomainPurchaser.sol
Remscar: To prevent variable shadowing. `_` is used for the parameter.

test/domainPurchaser.spec.ts
Remscar: We don't enforce naming or characters at a protocol level (smart contract level) since it's incredibly costly and almost impossible.

contracts/controllers/DomainPurchaser.sol
JamesEarle: Solidity will give errors about "shadowing" a variable if they share a name with one from a higher scope, this is just a way to make sure they're different but still "named" the same thing

contracts/controllers/DomainPurchaser.sol
JamesEarle: @Remscar this is the comment that should be changed because price don't necessarily need to be set

contracts/controllers/DomainPurchaser.sol
Remscar: In solidity all numbers are integers.

If I want to do the following: `360 * 0.5` it would be the same as `360 / 2`

Percentages are easier for humans to work with though, so we want to provide a way for users (owner in this case) to work with percentages.

By telling the user `1000` is one hundred percent we can then do the following;
```
divisionBasis = 1000; // 100%
percentFee = 10; // 1%
fee = (cost * percentFee) / divisonBasis
```
Which works out to `fee = (cost * 10) / 1000`

We could write this another way, which is more intuitive to us humans though:
`fee = cost * (10 / 1000)` (which is the same as) `fee = cost * (0.01)` 

However we need to keep in mind that solidity is only integers....
`10 / 1000 = 0` in solidity.
Therefore we first have to multiply the numerator, then do the divison, ie: `fee = (cost * 10) / 1000`

A division basis can be any number, and by making our division basis `1000` we are saying that a percentage can be at the smallest `0.1%` before it becomes zero.
If we made our basis `100` we could only go down to `1%` or with a basis of `10000` we could do `0.01%`

RyanCross: Got it, so division basis used to define what precision level the user can express a percentage. That might be handy as a comment if it can concisely fit.

contracts/controllers/DomainPurchaser.sol
RyanCross: A nice explanation was given during a 1:1 convo.

RyanCross (approved):

contracts/controllers/DomainPurchaser.sol
RyanCross: Got it. Thanks for the explanation of shadowing

PR #57, opened 2022-09-19
zNA Resolver contract

deep-quality-dev:
* Associate zNA to resource type
* Disassociate zNA from resource type
* Add resource registry for resource type


contracts/abstracts/ZeroUpgradeable.sol
```diff
@@ -0,0 +1,45 @@
...
+import {OwnableUpgradeable} from "../oz-upgradeable/access/OwnableUpgradeable.sol";
+import {AccessControlUpgradeable} from "../oz-upgradeable/access/AccessControlUpgradeable.sol";
+import {ReentrancyGuardUpgradeable} from "../oz-upgradeable/security/ReentrancyGuardUpgradeable.sol";
+import {PausableUpgradeable} from "../oz-upgradeable/security/PausableUpgradeable.sol";
+import {UUPSUpgradeable} from "../oz-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import {ContextUpgradeable} from "../oz-upgradeable/utils/ContextUpgradeable.sol";
+import {AddressUpgradeable} from "../oz-upgradeable/utils/AddressUpgradeable.sol";
+import {SafeMathUpgradeable} from "../oz-upgradeable/utils/math/SafeMathUpgradeable.sol";
+
+abstract contract ZeroUpgradeable is
```
Remscar: Don't use a base `ZeroUpgradeable` contract. This is more than just an upgradeable contract, it's pausable, has access control, ownable, and has reentrancy gaurds.

We don't 100% know we're going to need all this funcitonality in every zero contract which means we'll be wasting bytes.

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,262 @@
...
+pragma solidity ^0.8.11;
+
+// Uncomment this line to use console.log
+// import "hardhat/console.sol";
+import {ZeroUpgradeable} from "../abstracts/ZeroUpgradeable.sol";
+import {IResourceRegistry} from "../interfaces/IResourceRegistry.sol";
+import {IZNSHub} from "../interfaces/IZNSHub.sol";
+import {IZNAResolver} from "../interfaces/IZNAResolver.sol";
+
+contract ZNAResolver is ZeroUpgradeable, IZNAResolver {
```
Remscar: Realistically this only needs to derive from `AccessControl` and not `Ownable` also.

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,262 @@
...
+  uint256 public constant RESOURCE_TYPE_DAO = 0x1;
+  uint256 public constant RESOURCE_TYPE_STAKING_POOL = 0x2;
+  uint256 public constant RESOURCE_TYPE_FARMING = 0x4;
+
+  bytes32 public constant RESOURCE_TYPE_MANAGER_ROLE =
+    keccak256(abi.encode("RESOURCE_TYPE_MANAGER"));
+  bytes32 public constant RESOURCE_REGISTRY_MANAGER_ROLE =
+    keccak256(abi.encode("RESOURCE_TYPE_MANAGER"));
```
Remscar: Same hash in both of these constants, this should be RESOURCE_REGISTRY_MANAGER

deep-quality-dev: Thank you, I can fix it

contracts/resolver/ZNAResolver.sol
deep-quality-dev: I will update the parent contract.

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,262 @@
...
+
+  IZNSHub public znsHub;
+
+  // <zNA => ResourceType>
+  // ResourceTypes = DAO | StakingPool | Farming | ...
+  mapping(uint256 => uint256) private zNATypes;
```
Remscar: Might as well as make these mappings public. Easier to debug.

contracts/abstracts/ZeroUpgradeable.sol
deep-quality-dev: OK I will make it inherit from individual contracts we need

contracts/resolver/ZNAResolver.sol
deep-quality-dev: To do this, I made `resourceTypes` view function @Remscar

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,262 @@
...
+  mapping(uint256 => mapping(uint256 => uint256)) private zNAResourceIDs;
+
+  // <ResourceType, ResourceRegistry>
+  mapping(uint256 => IResourceRegistry) private resourceRegistries;
+
+  /* -------------------------------------------------------------------------- */
+  /*                                  Modifiers                                 */
+  /* -------------------------------------------------------------------------- */
+
+  modifier onlyResourceTypeManager(uint256 _zNA) {
```
Remscar: consideration: Potentially confusing modifier name since it's
sender is resource type manager OR owner of zNA

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,262 @@
...
+    require(
+      hasRole(RESOURCE_TYPE_MANAGER_ROLE, _msgSender()) ||
+        znsHub.ownerOf(_zNA) == _msgSender(),
+      "Not have a role"
```
Remscar: Improper grammer / not clear.

Consider 'not authorized: resource type manager'

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,262 @@
...
+    );
+    _;
+  }
+
+  modifier onlyResourceRegistryManager() {
+    require(
+      hasRole(RESOURCE_REGISTRY_MANAGER_ROLE, _msgSender()),
+      "Not have a role"
```
Remscar: Improper grammer

Consider 'not authorized: resource registry manager' or 'not authorized'

deep-quality-dev: OK Thanks

contracts/resolver/ZNAResolver.sol
deep-quality-dev: I will rename it `onlyResourceTypeManagerOrZNAOwner`

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,262 @@
...
+
+  /**
+   * @notice Update association of resource ID with resource type
+   * @dev Only callable by zNA Owner or the resource type manager
+   * @param _zNA Associating zNA
+   * @param _resourceType Single resource type. Check above constants.
+   * @param _resourceID Allocated resource ID. Resource ID can be zDAOId or
+   *     StakingPoolId or hash Id for other resource
+   */
+  function updateResourceIDWithResourceType(
```
Remscar: `associateWithResourceType` and `updateResourceIDWithResourceType` essentially do the same thing.
Having one function for set, and another function for update, is going to be confusing to users.

We should have only one function that sets/updates.

Combine the two functions, it is okay to have a branch statement to only call `zNATypes[_zNA] = zNATypes[_zNA] | _resourceType;` when needed

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,262 @@
...
+
+  /**
+   * @notice Check if resource type is valid. Only allow for pre-defined types
+   */
+  function _isValidResourceType(uint256 _resourceType)
+    internal
+    pure
+    returns (bool)
+  {
+    if (
```
Remscar: Can be simplified to
```
return (
      _resourceType == RESOURCE_TYPE_DAO ||
      _resourceType == RESOURCE_TYPE_STAKING_POOL ||
      _resourceType == RESOURCE_TYPE_FARMING
    )
```

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,262 @@
...
+   *     resource types by binary OR calculation.
+   *     If AND(current type, compare type) > 0, then return true.
+   */
+  function _hasResourceType(uint256 _zNA, uint256 _resourceType)
+    internal
+    view
+    returns (bool)
+  {
+    uint256 zNAType = zNATypes[_zNA];
+    if ((zNAType & _resourceType) > 0) return true;
```
Remscar: Can be simplified to `return (zNAType & _resourceType) > 0`

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,262 @@
...
+    return false;
+  }
+
+  /**
+   * @notice Check if resource exists by integrating with ResourceRegistry
+   */
+  function _isResourceExist(uint256 _resourceType, uint256 _resourceID)
```
Remscar: Improper grammer `doesResourceExist`

contracts/resolver/ZNAResolver.sol
deep-quality-dev: I received the same question from @RyanCross. Seems like not needed. I will fix it.

contracts/resolver/ZNAResolver.sol
deep-quality-dev: Great

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,262 @@
...
+  IZNSHub public znsHub;
+
+  // <zNA => ResourceType>
+  // ResourceTypes = DAO | StakingPool | Farming | ...
+  mapping(uint256 => uint256) private zNATypes;
+
+  // <zNA => <ResourceType => ResourceID>>
+  // ResourceID can be zDAOId or StakingPoolId or hash Id for other resource
+  // Single zNA can have multiple different resource
+  mapping(uint256 => mapping(uint256 => uint256)) private zNAResourceIDs;
```
Remscar: While we want to internally allow for a single zNA to have multiple resources of different types associated, we don't want to support that immediately.

Please update the contract to add logic that prevents more than one resource from being associated with a zNA. Don't change the storage layout, just change the logic in `associateWithResourceType`

deep-quality-dev: No problem

RyanCross:
@deep-quality-dev Ben and I had some good off PR conversations about the contract I want to share.  These were the main concerns brought up. (Note they are not necessarily in the scope of this initial story)

1. Should resolver keep track of `resourceIds`, and reuse them across the multiple ResourceRegistry contracts being created? E.g.

**Current**
image


VS.

image

2. Where, when, how is the resourceId generated? - TBD

3. How does a resource become associated to a type? - This process will be manual calling of the contract for now. 

3. Is there a need to be able to delete resource types? - TBD

contracts/abstracts/ZeroUpgradeable.sol
deep-quality-dev: I think using `ZeroUpgradeable` contract would be better, but we can remove unused contracts like `OwnableUpgradeable`, `ReentrancyGuardUpgradeable`, `PausableUpgradeable`.
If we're going to use `UUPS` pattern for upgradeable contract, the implementation contract should be inherited from `UUPSUpgradeable` and override `_authorizeUpgrade` (similar to ProxyAdmin).
The another good thing in `ZeroUpgrdeable` is `version`. 
I am going to check current version of contract through `version` storage variable. It means that when we upgrade to v2, then `version` should be 2.
I will add upgradeable test case more and let you know. Please review it then when you have time. @Remscar . Thanks

Remscar:
> @deep-quality-dev Ben and I had some good off PR conversations about the contract I want to share. These were the main concerns brought up. (Note they are not necessarily in the scope of this initial story)
> 
> 1. Should resolver keep track of `resourceIds`, and reuse them across the multiple ResourceRegistry contracts being created? E.g.
> 
> **Current** image
> 
> VS.
> 
> image
> 
> 2. Where, when, how is the resourceId generated? - TBD
> 3. How does a resource become associated to a type? - This process will be manual calling of the contract for now.
> 4. Is there a need to be able to delete resource types? - TBD

Resource ID's are supposed to be unique for each resource type.
A resource ID of 0x1 from type A should not be used in relation to type B.
ie; Type A, ID 1 != Type B, ID 1

Resource ID generation is the responsibility of the resource registry for that type. Each resource type needs it's own registry.

contracts/abstracts/ZeroUpgradeable.sol
deep-quality-dev: Removed `ZeroUpgradeable` and inherits only from `AccessControlUpgradeable`.

contracts/resolver/ZNAResolver.sol
deep-quality-dev: Now `ZNAResolver` only has `AccessControlUpgradeable` as parent

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,234 @@
...
+import {IResourceRegistry} from "../interfaces/IResourceRegistry.sol";
+import {IZNSHub} from "../interfaces/IZNSHub.sol";
+import {IZNAResolver} from "../interfaces/IZNAResolver.sol";
+
+contract ZNAResolver is AccessControlUpgradeable, IZNAResolver {
+  uint256 public constant RESOURCE_TYPE_DAO = 0x1;
+  uint256 public constant RESOURCE_TYPE_STAKING_POOL = 0x2;
+  uint256 public constant RESOURCE_TYPE_FARMING = 0x4;
+
+  bytes32 public constant RESOURCE_TYPE_MANAGER_ROLE =
```
Remscar, line 14: Can you please leave a comment detailing what each of these roles mean. What access do they give?

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * @notice Add resource registry with its resource type. Every resource type
+   *     can have a single resource registry.
+   *     Resource registry is responsible to check if resource exists here.
+   * @dev Only callable by resource registry manager
+   * @param _resourceType Single resource type. Check above constants
+   * @param _resourceRegistry Address to ResourceRegistry contract
+   */
+  function addResourceRegistry(
```
Remscar, line 131: I think i'm a bit confused on the `RESOURCE_TYPE_MANAGER` role.

I would assume that the `_resourceRegistry` would be given the `RESOURCE_TYPE_MANAGER` role in this call. And the old registry would lose that role.

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,234 @@
...
+    uint256 _resourceType,
+    uint256 _resourceID
+  ) external override onlyResourceTypeManagerOrZNAOwner(_zNA) {
+    require(znsHub.ownerOf(_zNA) != address(0), "Invalid zNA");
+    require(_isValidResourceType(_resourceType), "Invalid resource type");
+    require(
+      _doesResourceExist(_resourceType, _resourceID),
+      "Not exist resource"
+    );
+    require(
```
Remscar: Would be ideal if this would disassociate the existing resource instead of reverting.

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,234 @@
...
+    onlyResourceTypeManagerOrZNAOwner(_zNA)
+  {
+    require(znsHub.ownerOf(_zNA) != address(0), "Invalid zNA");
+    require(_isValidResourceType(_resourceType), "Invalid resource type");
+    require(_hasResourceType(_zNA, _resourceType), "Should have resource type");
+
+    zNATypes[_zNA] = zNATypes[_zNA] ^ _resourceType;
+    delete zNAResourceIDs[_zNA][_resourceType];
+
+    emit ResourceDisassociated(_zNA, _resourceType);
```
Remscar: I think ideally you would also emit the resource id of the resource that was disassociated. Just to be safe.

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,234 @@
...
+  // <ResourceType, ResourceRegistry>
+  mapping(uint256 => IResourceRegistry) public resourceRegistries;
+
+  /* -------------------------------------------------------------------------- */
+  /*                                  Modifiers                                 */
+  /* -------------------------------------------------------------------------- */
+
+  modifier onlyResourceTypeManagerOrZNAOwner(uint256 _zNA) {
+    require(
+      hasRole(RESOURCE_TYPE_MANAGER_ROLE, _msgSender()) ||
```
Remscar: A bit confused on who should have the `RESOURCE_TYPE_MANAGER_ROLE` role. I read this as "someone who manages resource types" but that doesn't seem to be how this role is being used.

This role is only ever assigned to the deployer/owner, and allows admin access to associate zNA's to resources. What is the goal of this role, is it meeting this goal, and is the name descriptive to the goal?

contracts/resolver/ZNAResolver.sol
deep-quality-dev: > Would be ideal if this would disassociate the existing resource instead of reverting.

I was going to write only two cases allowed.
- Set initial resource ID with a resource Type.
- Update existing resource ID with a resource type.
- Do not allow setting another resource Type on a zNA.

I will add more disassociating the existing resource.

contracts/resolver/ZNAResolver.sol
deep-quality-dev: > I think i'm a bit confused on the `RESOURCE_TYPE_MANAGER` role.
> 
> I would assume that the `_resourceRegistry` would be given the `RESOURCE_TYPE_MANAGER` role in this call. And the old registry would lose that role.

If we can say that `Resource Registry` have availability to associate its resource with zNA, then this function should give `RESOURCE_TYPE_MANAGER` to `Resource Registry`.

contracts/resolver/ZNAResolver.sol
deep-quality-dev: > I think ideally you would also emit the resource id of the resource that was disassociated. Just to be safe.

Sounds Great

contracts/resolver/ZNAResolver.sol
deep-quality-dev: Added the comment

deep-quality-dev: > Can you please leave a comment detailing what each of these roles mean. What access do they give?

Resource Type Manager is the role with who can associate zNA with resource type.
And Resource Registry Manager is the role with who can add resource registry to zNAResolver.

contracts/resolver/ZNAResolver.sol
deep-quality-dev: > A bit confused on who should have the `RESOURCE_TYPE_MANAGER_ROLE` role. I read this as "someone who manages resource types" but that doesn't seem to be how this role is being used.
> 
> This role is only ever assigned to the deployer/owner, and allows admin access to associate zNA's to resources. What is the goal of this role, is it meeting this goal, and is the name descriptive to the goal?

`zDAORegistry` is going to become ResourceRegistry. If we create new zDAO as Resource, then `zDAORegistry` will associate zNA and zDAO with `DAO` Resource Type.
So I think the following accounts can have `RESOURCE_TYPE_MANAGER_ROLE` role.
- `zDAORegistry`
- zNA owner
- owner of `zNAResolver` contract

Please let me know if I misunderstood.

Remscar: Understood, however it doesn't make sense for the `zNA Owner` to have the `RESOURCE_TYPE_MANAGER_ROLE` in that context.

contracts/resolver/ZNAResolver.sol
Remscar: This looks excellent now

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,285 @@
...
+
+  /**
+   * @notice Add resource registry with its resource type. Every resource type
+   *     can have a single resource registry.
+   *     Resource registry is responsible to check if resource exists here.
+   * @dev Only callable by resource registry manager
+   * @param _resourceType Single resource type. Check above constants
+   * @param _resourceRegistry Address to ResourceRegistry contract
+   */
+  function addResourceRegistry(
```
Remscar, line 131: Minor nit: Consider adding a require statement that will prevent `resourceRegistries[_resourceType]` from being overwritten with the same value. (require `resourceRegistries[_resourceType] != _resourceRegistry`)

contracts/resolver/ZNAResolver.sol
deep-quality-dev: > Understood, however it doesn't make sense for the `zNA Owner` to have the `RESOURCE_TYPE_MANAGER_ROLE` in that context.

Resource Registry can associate with creating new resource, so I thought that the another authorized account should be able to disassociate. zNAOwner can do this.

contracts/resolver/ZNAResolver.sol
deep-quality-dev: > Minor nit: Consider adding a require statement that will prevent `resourceRegistries[_resourceType]` from being overwritten with the same value. (require `resourceRegistries[_resourceType] != _resourceRegistry`)

Great

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,285 @@
...
+   * @notice Associate zNA with resource type.
+   *     Single zNA can have multiple different resource type and single
+   *     resource ID per each resource type.
+   * @dev Only callable by zNA Owner or the resource type manager
+   * @param _zNA Associating zNA
+   * @param _resourceType Single resource type. Check above constants.
+   * @param _resourceID Allocated resource ID. Resource ID can be zDAOId or
+   *     StakingPoolId or hash Id for other resource
+   */
+  function associateWithResourceType(
```
Remscar, line 85: There is a potential exploit in this architecture.

If this is called directly by the zNA owner, there is no exploit.

However; since this can be called by a "resource registry" which will associate a zNA to a resource *on behalf* of the user, there is the potential for an exploit.

If the resource registry does not verify that `_zNA` is owned by the initiator (user) of the transaction to the resource registry, it would be possible to create a resource and associate it to a zNA which the user does not own.

We are delegating the responsibility of ensuring that `_zNA` is owned by the user to the resource registry. If the resource registry fails to check this, we have an exploit vector.

I am of the belief that we shouldn't delegate the responsibility to the resource registry. It's easy for a downstream consumer to make a mistake. We should instead either;

- Create a 2nd function `associateWithResourceTypeOnBehalfOfUser` which can only be called by a resource registry. It takes an additional 4th parameter (`user`) which is the user we are associating the resource on behalf of. In this function we ensure that `user` owns `_zNA`
- OR: Add a 4th parameter to this function, `user` and ensure that `msg.sender === user` (unless msg.sender is a resource registry) and ensure that `user` owns `_zNA`

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,285 @@
...
+    uint256 _zNA,
+    uint256 _resourceType,
+    uint256 _resourceID
+  ) external override onlyResourceTypeManagerOrZNAOwner(_zNA) {
+    require(znsHub.ownerOf(_zNA) != address(0), "Invalid zNA");
```
Remscar: if you call `znsHub.ownerOf(_zNA)` with a `_zNA` that does not exist, a revert will throw; https://github.com/zer0-os/zNS/blob/18f416c188b771fc4dc54a27e069d571ef90090e/contracts/registrar/ZNSHub.sol#L107

You are looking for `domainExists` i believe

deep-quality-dev: > 

Yes, I need validation check for zNA, I can update it. Thanks for letting me know.

contracts/resolver/ZNAResolver.sol
deep-quality-dev: > There is a potential exploit in this architecture.
> 
> If this is called directly by the zNA owner, there is no exploit.
> 
> However; since this can be called by a "resource registry" which will associate a zNA to a resource _on behalf_ of the user, there is the potential for an exploit.
> 
> If the resource registry does not verify that `_zNA` is owned by the initiator (user) of the transaction to the resource registry, it would be possible to create a resource and associate it to a zNA which the user does not own.
> 
> We are delegating the responsibility of ensuring that `_zNA` is owned by the user to the resource registry. If the resource registry fails to check this, we have an exploit vector.
> 
> I am of the belief that we shouldn't delegate the responsibility to the resource registry. It's easy for a downstream consumer to make a mistake. We should instead either;
> 
> * Create a 2nd function `associateWithResourceTypeOnBehalfOfUser` which can only be called by a resource registry. It takes an additional 4th parameter (`user`) which is the user we are associating the resource on behalf of. In this function we ensure that `user` owns `_zNA`
> * Add a 4th parameter to this function, `user` and ensure that `msg.sender === user` (unless msg.sender is a resource registry) and ensure that `user` owns `_zNA`

Would you please review with this PR together?
https://github.com/zer0-os/zdao-contracts/pull/26
I can understand the potential risk of this function, so I rolled back `onlyZNAOwner` in `addNewZDAO` from `zDAORegistry`. It means that only zNA owner can add new zDAO and then `zDAORegistry` will associate it.
Associated zNA can be disassociated by zNAOwner or any accounts which have the role of RESOURCE_TYPE_MANAGER.

Remscar: Okay that is good you do the check there as well.
I am thinking it may be more secure to do the check here.

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,285 @@
...
+  function _associateWithResourceType(
+    uint256 _zNA,
+    uint256 _resourceType,
+    uint256 _resourceID
+  ) internal {
+    if (_hasResourceType(_zNA, _resourceType)) {
+      _disassociateWithResourceType(_zNA, _resourceType);
+    } else if (zNATypes[_zNA] > 0) {
+      // At this moment, do not support multiple resource type per a zNA
+      revert("Only support single resource type");
```
Remscar: This should disassociate the other resource type instead of reverting.

It is not intuitive that if the resource is already associated with DAO #1 and i want to associate it with DAO #2 that it would not revert. But if it was associated with Staking Pool #1 and I want to associate it with DAO #2, then it would revert.

The behavior should be the same across similar use cases.

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,285 @@
...
+   * @param _zNA Associating zNA
+   * @param _resourceType Single resource type. Check above constants.
+   * @param _resourceID Allocated resource ID. Resource ID can be zDAOId or
+   *     StakingPoolId or hash Id for other resource
+   */
+  function associateWithResourceType(
+    uint256 _zNA,
+    uint256 _resourceType,
+    uint256 _resourceID
+  ) external override onlyResourceTypeManagerOrZNAOwner(_zNA) {
```
Remscar: There is a potential bug / exploit in this function;

Since this function can be called by either the `zNA Owner` or a `Resource Registry` we need to perform some additional safety checks in the case of a `Resource Registry`

If the `Resource Registry` is associating a zNA to a resource, on behalf of a user, then the resource registry should only be able to associate resources whose type it manages. For instance, the zDAO Registry (The resource registry for zDAO's) should only be able to associate zNA's to zDAO's. It should not be able to associate zNA's to something like Staking Pools. Similarly the Staking Pool Registry should only be able to associate Staking Pools.

If the caller of this function is a resource registry, you need to verify that the `resourceRegistries[_resourceType] === msg.sender`.

This will prevent potential bugs and attacks where a resource registry may register a resource of a type that it doesn't actually control.

scripts/resolver/deploy-znaresolver.ts
```diff
@@ -0,0 +1,62 @@
...
+    network.name === "rinkeby" ||
+    network.name === "mainnet"
+  ) {
+    console.log("Deploying ZNAResolver proxy contract...");
+    const ZNAResolverFactory = await ethers.getContractFactory("ZNAResolver");
+    const zNAResolver = (await upgrades.deployProxy(
+      ZNAResolverFactory,
+      [config[network.name].zNSHub],
+      {
+        kind: "uups",
```
Remscar: This seems liket he wrong kind. The zNA resolver seems to require a transparent proxy, not UUPS, since there is no `upgradeTo` functions

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,285 @@
...
+// import "hardhat/console.sol";
+import {AccessControlUpgradeable} from "../oz442/access/AccessControlUpgradeable.sol";
+import {IResourceRegistry} from "../interfaces/IResourceRegistry.sol";
+import {IZNSHub} from "../interfaces/IZNSHub.sol";
+import {IZNAResolver} from "../interfaces/IZNAResolver.sol";
+
+contract ZNAResolver is AccessControlUpgradeable, IZNAResolver {
+  uint256 public constant RESOURCE_TYPE_DAO = 0x1;
+  uint256 public constant RESOURCE_TYPE_STAKING_POOL = 0x2;
+  uint256 public constant RESOURCE_TYPE_FARMING = 0x4;
```
Remscar: Is it possible for us to move these into a library, or perhaps add a gap for upgradeability.
I'm thinking about how we're going to add additional types in the future and we want to make this as easy as possible.

contracts/resolver/ZNAResolver.sol
```diff
@@ -0,0 +1,285 @@
+// SPDX-License-Identifier: UNLICENSED
```
Remscar: MIT license.

deep-quality-dev: Updated

contracts/resolver/ZNAResolver.sol
deep-quality-dev: Thanks for your checking

scripts/resolver/deploy-znaresolver.ts
deep-quality-dev: Updated deploy script with transparent pattern

contracts/resolver/ZNAResolver.sol
deep-quality-dev: Constant variables are not reserved storage slots.
> https://docs.soliditylang.org/en/v0.8.16/contracts.html#constant-and-immutable-state-variables

So I think these will not affect upgrading to the new version.

contracts/resolver/ZNAResolver.sol
deep-quality-dev: Great point, indeed. Resource Registry should be checked with its registered resource type.

contracts/resolver/ZNAResolver.sol
deep-quality-dev: Sounds Great.
Actually, that's what I originally thought.
Now Resource Registry is checked with resource type, then making this function to allow re-associating with different resource type would be perfect.

contracts/resolver/ZNAResolver.sol
deep-quality-dev: I am going to move those constants to the new library and import it in `ZDAORegistry`, too.

contracts/resolver/ZNAResolver.sol
deep-quality-dev: Please check this PR: https://github.com/zer0-os/zNS/pull/58

contracts/resolver/ZNAResolver.sol
deep-quality-dev: Please check this PR: https://github.com/zer0-os/zNS/pull/58

contracts/resolver/ZNAResolver.sol
deep-quality-dev: Please check this PR: https://github.com/zer0-os/zNS/pull/58

Remscar (approved):

PR #58, opened 2022-10-04
Fix resource type authorization

deep-quality-dev:
* use `domainExists` to check if zNA exists
* revoke `RESOURCE_TYPE_MANAGER_ROLE` role from old resource registry when add new one
* disassociate zNA if associate with different resource type
* only allow to manage registered resource type: DAO resource registry can't associate with Staking resource type
* only zNA owner can disassociate zNA with resource type: resource registry can't disassociate when delete resource itself
* use Resource Type constants as library
* update unit test

### DIFF
--- a/test/zNAResolver.spec.ts
+++ b/test/zNAResolver.spec.ts
@@ -6,54 +6,76 @@ import {
 } from "@defi-wonderland/smock";
 import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
 import chai, { expect } from "chai";
-import { BigNumber } from "ethers";
+import { BigNumber, ContractTransaction } from "ethers";
 import { ethers } from "hardhat";
 import {
-  IResourceRegistry,
   IZNSHub,
   ZNAResolver,
   ZNAResolver__factory,
 } from "../typechain";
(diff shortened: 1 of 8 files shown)


test/zNAResolver.spec.ts
```diff
@@ -6,84 +6,193 @@ import {
...
-    )) as FakeContract<IResourceRegistry>;
+    // Deploy ResourceType Library
+    const ResourceTypeLibFactory = await ethers.getContractFactory("ResourceType");
+    resourceTypeLib = await ResourceTypeLibFactory.deploy() as ResourceType;
 
+    // What is Resource Type
+    RESOURCE_TYPE_DAO = await resourceTypeLib.RESOURCE_TYPE_DAO();
+    RESOURCE_TYPE_STAKING_POOL = await resourceTypeLib.RESOURCE_TYPE_STAKING_POOL();
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    RESOURCE_TYPE_FARMING = await resourceTypeLib.RESOURCE_TYPE_FARMING();
```
Remscar: doesn't exist anymore?

Remscar (approved):

PR #60, opened 2022-10-27
Scripts for root migration

JamesEarle:
[Steps noted here](https://www.notion.so/zerotech/Root-Domain-Migration-14b83742575740a2a3b7add1f24e20c0)

Perform a simulated root migration in hardhat. Also, when ready test in Goerli, confirm it all works correctly and then run again on mainnet.

Closes [AB#1063](https://dev.azure.com/zero-wilderworld/07337eb7-a009-4103-8b97-ecb872053d7e/_workitems/edit/1063)


contracts/oz/token/ERC721/ERC721Upgradeable.sol
```diff
@@ -506,7 +506,8 @@ contract ERC721Upgradeable is
       ERC721Upgradeable.ownerOf(tokenId) == from,
       "ERC721: transfer of token that is not own"
     ); // internal owner
-    require(to != address(0), "ERC721: transfer to the zero address");
```
Remscar, line 509: Would be good idea to rename this file with this change, so you have the old version still intact and can easily switch back to if needed.

contracts/registrar/ZNSHub.sol
```diff
@@ -91,8 +91,6 @@ contract ZNSHub is
   // Beacon Proxy used by zNS Registrars
   address public beacon;
 
-  uint8 private test;
```
Remscar: You can't remove variables from smart contracts once they're already there, unless you use special flags in the upgrade process

scripts/rootMigration/actions/runSimulation.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  await zNSHub.addRegistrar(ethers.constants.HashZero, legacyRegistrar.address);
+  await legacyRegistrar.connect(signer).addController(signerAddress);
+
+  const [rootDomainId, wilderDomainId] = await mintSamples(signer, legacyRegistrar);
+
+  // Begin Migration simulation
+  // 2. Upgrade existing proxies to have needed functionality
+  // 2.a. Upgrade legacy Registrar
+  const upgradeRegistrarTx: MigrationRegistrar = await hre.upgrades.upgradeProxy(
```
Remscar: You should be running the same code to migrate and simulate. Refactor the business logic from `runMigration` into a utility file, then import and run that code in `runSimulation`. If the code isn't the same, it's not a true simulation and you could have a bug in one, but not the other.

scripts/rootMigration/actions/runSimulation.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  // 2.b. Upgrade Beacon registrars
+  const upgradeBeaconTx: MigrationRegistrar = await hre.upgrades.upgradeBeacon(
+    beacon,
+    new MigrationRegistrar__factory(signer)
+  ) as MigrationRegistrar;
+  const upgradedBeacon = await upgradeBeaconTx.deployed();
+
+  // 3. Burn domains
+  await upgradedLegacyRegistrar.connect(signer).burnDomain(wilderDomainId);
```
Remscar: You are going to want to `.wait(2)` on any and all transactions you do. When running in hardhat, you will want to only do `.wait()`

scripts/rootMigration/actions/runSimulation.ts
```diff
@@ -0,0 +1,173 @@
...
+    beaconRegistrar.address
+  );
+
+  const newWilderDomainId = await getDomainId(tx);
+
+  console.log("new wilder domain: ", newWilderDomainId.toHexString());
+
+  let tx_set;
+  // 6. Update the rootDomainId and parentRegistrar values on upgraded legacy
+  tx_set = await upgradedLegacyRegistrar.connect(signer).setRootDomainId(ethers.constants.HashZero);
```
Remscar: Consider just making 'set root domain id' and 'set parent registrar' to be the same function

scripts/rootMigration/actions/runSimulation.ts
```diff
@@ -0,0 +1,173 @@
...
+  const beaconRegistrar = await newRegistrarTx.deployed();
+
+  console.log(`beaconRegistrar: ${beaconRegistrar.address}`)
+
+  await zNSHub.addRegistrar(rootDomainId, beaconRegistrar.address);
+
+  // 5. Mint 0://wilder in new root registrar
+  const tx = await beaconRegistrar.connect(signer).mintDomain(
+    ethers.constants.HashZero, // parent id
+    "0://wilder",
```
Remscar: this should just be `wilder`

We don't include 0:// when minting domains, just the label

scripts/rootMigration/postMigrate.ts
```diff
@@ -0,0 +1,25 @@
...
+//    0a. Make this a new version, maintain the old version, PM, for post migration
+// 1. Upgrade registrars with required code changes
+// 2. Burn 0:// and 0://wilder on legacy registrar
+// 3. Deploy new root registrar (beacon proxy)
+// 4. Mint 0://wilder (we will not remint the root 0://)
+// 5. Update `rootDomainId` and `parentRegistrar` on legacy registrar
+//    5a. Make these values point to the newly minted 0://wilder domain on beacon
+// 6. Upgrade both the legacy registrar and new beacon to have the old registrar code (PM)
+
+const migrate = async () => {
```
Remscar: What's the point of this script? It looks like `runSimuation` already does this

scripts/rootMigration/actions/runSimulation.ts
```diff
@@ -0,0 +1,173 @@
...
+    false,
+    signerAddress,
+    beaconRegistrar.address
+  );
+
+  const newWilderDomainId = await getDomainId(tx);
+
+  console.log("new wilder domain: ", newWilderDomainId.toHexString());
+
+  let tx_set;
```
Remscar: ultra nit pick: python naming convention on this variable

scripts/rootMigration/actions/runSimulation.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  // 5. Mint 0://wilder in new root registrar
+  const tx = await beaconRegistrar.connect(signer).mintDomain(
+    ethers.constants.HashZero, // parent id
+    "0://wilder",
+    signerAddress,
+    "ipfs.io/ipfs/Qm",
+    0,
+    false,
+    signerAddress,
```
Remscar: Your `minter` and `sendToUser` are the same. Remove the send to user logic if you're not sending it to a user other than the minter

Remscar (changes requested):
Migration and simulation should be the exact same code, not copied code. It's dangerous to have them be different code since you might fix a bug in one but forget to fix a bug in the other. Also you will need to make sure you `.wait()` every transaction otherwise you'll get a failure. I would also recommend waiting for atleast 2-3 confirmations on mainnet.

If you want to make this safer, consider writing it so that it can be re-run without repeating steps in case of a failure mid-way through. For instance if we crash after upgrading the proxy beacon, let's just skip over that step. Might be much harder to code this, but could help prevent a sticky situation where you need to comment out code and re-run the script because of an transient error during the actual migration.

contracts/oz/token/ERC721/ERC721Upgradeable.sol
JamesEarle: Good idea

scripts/rootMigration/postMigrate.ts
JamesEarle: meant to delete this, I had several before

contracts/oz/token/ERC721/ERC721PausableUpgradeable.sol
```diff
@@ -2,7 +2,7 @@
 
 pragma solidity ^0.8.9;
 
-import "./ERC721Upgradeable.sol";
+import "./ERC721UpgradeableBurnable.sol";
```
Remscar, line 5: Would be good to rename this file as well, like you did with UpgradeableBurnable.

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,143 @@
...
+    [
+      hre.ethers.constants.AddressZero, // Parent registrar
+      rootDomainId,
+      "Zer0 Namespace Service",
+      "ZNS",
+      zNSHub.address
+    ]
+  );
+
+  const beaconRegistrar = await newRegistrarTx.deployed() as MigrationRegistrar;
```
Remscar: Not contextual variable name. I'm assuming this is the new (0://) registrar?

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,143 @@
...
+  if (wilderDomainExists || rootDomainExists) {
+    throw Error("Burn didn't work");
+  }
+
+  logger.log("4. Deploy new root registrar as proxy to existing subdomain registrar beacon");
+  const newRegistrarTx = await hre.upgrades.deployBeaconProxy(
+    upgradedBeacon.address,
+    new MigrationRegistrar__factory(signer),
+    [
+      hre.ethers.constants.AddressZero, // Parent registrar
```
Remscar: AddressZero technically is not the same as HashZero. This works, but it would be more correct to use HashZero.

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,143 @@
...
+    const [zNSHub, legacyRegistrar, beacon] = await helpers.deployContracts();
+
+    await zNSHub.addRegistrar(hre.ethers.constants.HashZero, legacyRegistrar.address);
+    await legacyRegistrar.connect(signer).addController(signerAddress);
+
+    legacyRegistrarAddress = legacyRegistrar.address;
+    zNSHubAddress = zNSHub.address;
+    beaconAddress = beacon.address;
+
+    wilderDomainId = await helpers.mintSampleWilderDomain(signer, legacyRegistrar);
```
Remscar: Why would you not use the `wilder` domain and instead use `wilder-sample` if we're doing a hardhat fork?

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,143 @@
...
+  let zNSHubAddress;
+  let beaconAddress;
+  let wilderDomainId;
+
+  const rootDomainId = "0";
+  const waitBlocks = network === "hardhat" ? 0 : 3;
+
+  if (network === "hardhat") {
+    // Setup default contracts required for migration simulation
+    const [zNSHub, legacyRegistrar, beacon] = await helpers.deployContracts();
```
Remscar: Why redeploy the contracts instead of use the existing contracts on mainnet? The hardhat network is a local fork, and can be configured to fork from mainnet, allowing for a true simulation.

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,143 @@
...
+      zNSHub.address
+    ]
+  );
+
+  const beaconRegistrar = await newRegistrarTx.deployed() as MigrationRegistrar;
+
+  logger.log(`New Beacon Registrar Address: ${beaconRegistrar.address}`);
+
+  // Must do this to be able to mint below
+  await zNSHub.addRegistrar(rootDomainId, beaconRegistrar.address);
```
Remscar: You are not waiting for this tx to complete

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,143 @@
...
+  logger.log("Verify burn");
+  const wilderDomainExists = await upgradedLegacyRegistrar.domainExists(wilderDomainId);
+  const rootDomainExists = await upgradedLegacyRegistrar.domainExists(rootDomainId);
+
+  if (wilderDomainExists || rootDomainExists) {
+    throw Error("Burn didn't work");
+  }
+
+  logger.log("4. Deploy new root registrar as proxy to existing subdomain registrar beacon");
+  const newRegistrarTx = await hre.upgrades.deployBeaconProxy(
```
Remscar: You are not waiting for this tx to complete

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  const beaconRegistrar = await newRegistrarTx.deployed() as MigrationRegistrar;
+
+  logger.log(`New Beacon Registrar Address: ${beaconRegistrar.address}`);
+
+  // Must do this to be able to mint below
+  await zNSHub.addRegistrar(rootDomainId, beaconRegistrar.address);
+
+  logger.log("5. Mint wilder in new root registrar");
+  const tx = await beaconRegistrar.connect(signer).mintDomain(
```
Remscar: You are not waiting for this tx to complete

Remscar (changes requested):
You're still not waiting for all transactions to have block confirmations.
If you use libraries like `hre.upgrades` you need to figure out how to get the tx receipt and how to wait for block confirmations with those as well.

scripts/rootMigration/actions/runMigration.ts
JamesEarle: yes, changed the name to `newRootRegistrar` to be clearer

scripts/rootMigration/actions/runMigration.ts
JamesEarle: it was wilder before, I modified it to confirm the issue I DMd you about updating subdomain parents that wasn't an issue and just didn't change it back

scripts/rootMigration/actions/runMigration.ts
JamesEarle: It is being waited in the following function call to `getDomainId` but because that is no longer getting used more than once i've brought the wait directly in here

JamesEarle:
> You're still not waiting for all transactions to have block confirmations. If you use libraries like `hre.upgrades` you need to figure out how to get the tx receipt and how to wait for block confirmations with those as well.

For each use of a HardHat function for upgrading or deploying something we call `await newThing.deployed()` which internally calls to wait on the transaction then gives you the contract from there.
image

contracts/registrar/MigrationRegistrar.sol
```diff
@@ -231,8 +229,12 @@ contract MigrationRegistrar is
   // vv TEMPORARY FOR MIGRATION vv
   //
 
-  function burnDomain(uint256 domainId) external onlyOwner {
-    _transfer(msg.sender, address(0), domainId);
+  function burnDomains(
```
Remscar, line 232: I have a slight worry about a function like this. It shouldn't pose an issue but imagine if something goes wrong and you need to burn only a single domain. This function must burn two at a time; you can't burn just one domain.

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,184 @@
...
+  logger.log("5. Mint wilder in new root registrar");
+  confirmContinue();
+  const tx = await newBeaconRegistrar.connect(signer).mintDomain(
+    hre.ethers.constants.HashZero, // Parent domainId
+    "wilder",
+    signerAddress,
+    "ipfs://QmSQTLzzsPS67ay4SFKMv9Dq57iSQ7pLWQVUvS3XB5MowK/0",
+    0,
+    false,
+    newBeaconRegistrar.address
```
Remscar: This is the wrong address. This should be the address of the wilder registrar (legacyRegistrarAddress)

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  logger.log("2. Upgrade existing proxies to have needed functionality");
+
+  logger.log("2.a. Upgrade legacy Registrar");
+  confirmContinue();
+  const upgradeRegistrarTx = await hre.upgrades.upgradeProxy(
+    legacyRegistrarAddress,
+    new MigrationRegistrar__factory(signer)
+  );
+  const upgradedLegacyRegistrar = await upgradeRegistrarTx.deployed() as MigrationRegistrar;
```
Remscar: You are still not properly waiting for these transactions to be confirmed. `.deployed()` waits only 1 block, you should wait `waitBlock` number of blocks by doing `upgradeRegistrarTx.deployTransaction.wait(waitBlocks)`

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  logger.log("2.b. Upgrade Beacon registrars");
+  confirmContinue();
+  const upgradeBeaconTx = await hre.upgrades.upgradeBeacon(
+    beaconAddress,
+    new MigrationRegistrar__factory(signer)
+  );
+  // Beacon is the subdomain contract for wolf
+  const upgradedLegacyRegistrarBeacon = await upgradeBeaconTx.deployed() as MigrationRegistrar;
```
Remscar: This is an invalid cast, this is not a `MigrationRegistrar` as you have casted it. It is a `ProxyBeacon` contract

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,184 @@
...
+  );
+  const upgradedLegacyRegistrar = await upgradeRegistrarTx.deployed() as MigrationRegistrar;
+
+  logger.log("2.b. Upgrade Beacon registrars");
+  confirmContinue();
+  const upgradeBeaconTx = await hre.upgrades.upgradeBeacon(
+    beaconAddress,
+    new MigrationRegistrar__factory(signer)
+  );
+  // Beacon is the subdomain contract for wolf
```
Remscar, line 71: ?

scripts/rootMigration/actions/runMigration.ts
JamesEarle: Interesting, I expected that tx to return the type of contract we were upgrading, like `upgradeProxy` does but you're right. We don't use it so it didn't come up but good catch.

contracts/registrar/MigrationRegistrar.sol
```diff
@@ -0,0 +1,810 @@
...
+  function lockDomainMetadata(uint256 id, bool toLock) external override {
+    _validateLockDomainMetadata(id, toLock);
+    _setDomainLock(id, msg.sender, toLock);
+  }
+
+  /*
+   * Public View
+   */
+
+  // function ownerOf(uint256 tokenId)
```
Whytecrowe, line 481: Why is this commented out?

contracts/registrar/MigrationRegistrar.sol
```diff
@@ -0,0 +1,810 @@
...
+      keccak256(abi.encodePacked(domainGroups[id].baseMetadataUri)) !=
+        keccak256(abi.encodePacked(baseMetadataUri)),
+      "Folder groups are the same"
+    );
+    domainGroups[id].baseMetadataUri = baseMetadataUri;
+
+    zNSHub.domainGroupUpdated(id, baseMetadataUri);
+  }
+
+  modifier onlyController() {
```
Whytecrowe, line 104: We seem to have 2 "functions" that check "if controller". This and `isController()` below. This seem to allow ZNSHub contract's controller to call, but not in `isController()`. It might be needed for the migration only and disallow the ZNSHub's controller to do it normally.
Just wanted to double-check to make sure both of them are needed.

contracts/registrar/MigrationRegistrar.sol
```diff
@@ -0,0 +1,810 @@
...
+    return accountIsController;
+  }
+
+  /**
+   * @notice Returns whether or not a domain is exists
+   * @param id The domain
+   */
+  function domainExists(uint256 id) public view override returns (bool) {
+    bool domainNftExists = _exists(id);
+    bool domainHasNotBeenBurned = ownerOf(id) != address(0);
```
Whytecrowe, line 512: This seems to use the function commented out above. Maybe just a temp thing while working on the PR...

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,183 @@
...
+  // hardhat to be able to allow us to upgrade it.
+  if (network === "hardhat") {
+    await hre.upgrades.forceImport(legacyRegistrarAddress, new Registrar__factory());
+  }
+
+  const zNSHub = ZNSHub__factory.connect(zNSHubAddress, signer);
+
+  logger.log("Comparing bytecode of Registrar and MigrationRegistrar for compatibility");
+  logger.log(`Registrar Bytecode: ${new Registrar__factory(signer).bytecode.length / 2}`);
+  logger.log(`MigrationRegistrar Bytecode: ${new MigrationRegistrar__factory(signer).bytecode.length / 2}`);
```
Whytecrowe: Why do we divide by 2 here?

scripts/rootMigration/actions/runMigration.ts
```diff
@@ -0,0 +1,183 @@
...
+  signer: SignerWithAddress,
+  network: string
+) => {
+  const signerAddress = await signer.getAddress();
+  const addresses = getAddressesForNetwork(network);
+
+  logger.log(`Using network ${network} for migration`);
+
+  const rootDomainId = "0";
+  const waitBlocks = network === "hardhat" ? 0 : 3;
```
Whytecrowe, line 36: why we wait for 3 blocks?

Whytecrowe:
> If you want to make this safer, consider writing it so that it can be re-run without repeating steps in case of a failure mid-way through. For instance if we crash after upgrading the proxy beacon, let's just skip over that step. Might be much harder to code this, but could help prevent a sticky situation where you need to comment out code and re-run the script because of an transient error during the actual migration.

Second this! If possible, should look into it!

scripts/rootMigration/migrate.ts
```diff
@@ -0,0 +1,21 @@
...
+// 4. Mint 0://wilder (we will not remint the root 0://)
+// 5. Update `rootDomainId` and `parentRegistrar` on legacy registrar
+//    5a. Make these values point to the newly minted 0://wilder domain on beacon
+// 6. Upgrade both the legacy registrar and new beacon to have the old registrar code (PM)
+
+const migrate = async () => {
+  await hre.run("compile");
+
+  const [signer] = await hre.ethers.getSigners();
+  runMigration(signer, hre.network.name);
```
Whytecrowe: shouldn't this be awaited or returned?

contracts/oz/token/ERC721/ERC721Upgradeable.sol
```diff
@@ -506,7 +506,8 @@ contract ERC721Upgradeable is
...
+    // To allow token burning we must be able to send to the zero address
+    // require(to != address(0), "ERC721: transfer to the zero address");
```
deep-quality-dev, line 510: Burning token calls `_burn` function in `ERC721Upgradeable` contract, it will not call `_transfer` anymore.
We don't need to comment this `require` for burnning.

contracts/registrar/MigrationRegistrar.sol
```diff
@@ -0,0 +1,816 @@
...
+
+  function burnDomains(
+    uint256 _wilderDomainId,
+    uint256 _rootDomainId,
+    address _wilderOwner,
+    address _rootOwner
+  ) external onlyOwner {
+    // ERC721 transfer will check owner internally
+    _transfer(_wilderOwner, address(0), _wilderDomainId);
+    _transfer(_rootOwner, address(0), _rootDomainId);
```
deep-quality-dev, line 240: Burning should use `_burn` function instead of `_transfer`

scripts/rootMigration/actions/runMigrationFromSafe.ts
```diff
@@ -0,0 +1,426 @@
...
+  /**
+   * Call rootRegistrar.mintDomain(parentDomainId,label,minter,metadataUri,royaltyAmount,locked,subdomainContract)
+   * 
+   * Will mint new domain with as `label` under `parentDomain`
+   */
+  await createAndProposeTransaction(
+    safeAddress,
+    signer,
+    rootRegistrar,
+    "mintDomain(uint256,string,address,string,uint256,bool,address)",
```
deep-quality-dev, line 354: Before creating tx proposal to Gnosis Safe, `rootRegistrar` should transfer ownership to Gnosis Safe.
By default, [`hre.upgrades.deployBeaconProxy`](https://github.com/zer0-os/zNS/pull/60/files#diff-b43767899798227726329bb33b27e41ae4cd22276087ced4da127b52bf1f74dfR299) use deployer EOA as the owner of contract.

scripts/rootMigration/actions/runMigrationFromSafe.ts
```diff
@@ -0,0 +1,426 @@
...
+  const nonce = await safeService.getNextNonce(safeSdk.getAddress())
+
+  // Don't add nonce while testing because you can't run simulations on txs that it knows
+  // have to follow others based on their nonce
+  const safeTransactionData: SafeTransactionDataPartial = {
+    to: to,
+    data: data,
+    value: "0",
+    gasPrice: 1000000000, // 1 gwei
+    safeTxGas: 60000,
```
deep-quality-dev, line 123: Is this gas fee enough to execute safe transaction?
I think it should be set by 0, so it can be reverted if internal tx is failed and use enough gas fees.
Please refer this link: https://github.com/safe-global/safe-contracts/blob/main/docs/safe_tx_gas.md

PR #61, opened 2022-11-17
Feature/opensea operator filter

beardhero:
Incorporates opensea operator filter functionality with the Registrar


JamesEarle (approved):

JamesEarle (changes requested):
Some things pending

contracts/opensea/DefaultOperatorFilterer.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
```
JamesEarle: This file looks unused because you use the `OperatorFilterer` contract direct just calling to `initializeFilter` with the same address in the Registrar. Please remove unused code.

contracts/opensea/OwnedRegistrant.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
```
JamesEarle: This file is unused as well, and also is using a constructor and wouldn't support upgradability modifications you made to the others. Please remove.

contracts/oz/access/Ownable2Step.sol
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v4.7.0) (access/Ownable.sol)
+
+pragma solidity ^0.8.0;
```
JamesEarle: If this file is only being used by the above, unused, `OwnedRegistrant` please also remove this.

contracts/registrar/Registrar.sol
```diff
@@ -57,8 +59,6 @@ contract Registrar is
 
   // The event emitter
   IZNSHub public zNSHub;
-  uint8 private test; // ignore
```
JamesEarle: When I tested using my own deployment script these changes didn't throw errors but alterations to storage layouts like this and the above addition of the `contract Registrar is ..., OperatorFilter` would be an unsafe upgrade. Please write your own upgrade script to confirm that these changes are upgrade safe.

contracts/registrar/Registrar.sol
```diff
@@ -476,31 +486,20 @@ contract Registrar is
...
-  }
-
   /**
    * @notice Returns whether or not a domain is exists
    * @param id The domain
    */
   function domainExists(uint256 id) public view override returns (bool) {
-    bool domainNftExists = _exists(id);
-    return domainNftExists;
+    return _exists(id);
```
JamesEarle, line 502: Even with simpler functions like this code readability and debug ability is improved when we first assign to a variable and return it. Considering this change is already made by the optimizer when we deploy the contract I'd recommend you revert this and the several other functions below it.

contracts/registrar/Registrar.sol
```diff
@@ -681,7 +674,7 @@ contract Registrar is
     }
   }
 
-  /**
+  /**`
```
JamesEarle: Remove backtick

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "devDependencies": {
-    "@defi-wonderland/smock": "2.0.7",
+    "@defi-wonderland/smock": "2.3.0",
     "@nomiclabs/ethereumjs-vm": "4.2.2",
-    "@nomiclabs/hardhat-ethers": "2.0.5",
+    "@nomiclabs/hardhat-ethers": "^2.2.1",
```
JamesEarle: Please remove `^` here and elsewhere

hardhat.config.ts
```diff
@@ -1,5 +1,7 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 require("dotenv").config();
+//require('@eth-optimism/smock/build/src/plugins/hardhat-storagelayout')
```
JamesEarle: Remove unused code

hardhat.config.ts
```diff
@@ -81,7 +97,7 @@ const config: HardhatUserConfig = {
       //   },
       // ],
       forking: {
-        url: "https://rinkeby.infura.io/v3/fa959ead3761429bafa6995a4b25397e",
+        url: "https://eth-mainnet.alchemyapi.io/v2/-nhhIZg46QlTmzPozXF07vyxpK5BGukx",
```
JamesEarle: Did changing from `rinkeby` to mainnet but still using `infura` not work? I'd avoid changing providers

package.json
```diff
@@ -53,5 +53,8 @@
   },
   "name": "zns-contracts",
   "version": "0.0.0",
-  "dependencies": {}
-}
\ No newline at end of file
+  "dependencies": {
+    "@eth-optimism/smock": "^1.1.10",
+    "forge-std": "^1.1.2"
```
JamesEarle: Seems like neither if these are used (the smock package is commented out in the hardhat config to boot) please remove unused code

contracts/registrar/Registrar.sol
```diff
@@ -140,6 +148,8 @@ contract Registrar is
 
     __ERC721Pausable_init();
     __ERC721_init(collectionName, collectionSymbol);
+
+    initializeFilter(address(0x3cc6CddA760b79bAfa08dF41ECFA224f810dCeB6), true);
```
JamesEarle: I'm realizing this won't work for anything except this registrar and registrar's going forwards in time, but we have several existing registrars that will also need this update and won't receive it via this upgrade. Please add functionality to set this, likely through the hub as it points to the already existing registrars in question.

JamesEarle (commented):
Thanks for the tests. There are a few things I didn't think about on my earlier pass through, so please address each. Also, I mentioned I did a small test upgrade script myself and it seemed okay but please write your own to make sure it upgrades the existing registrar and proxy beacon to make sure all the upgrade comments I made are safe.

hardhat.config.ts
beardhero: removed

contracts/registrar/Registrar.sol
beardhero: removed

contracts/registrar/Registrar.sol
beardhero: This is not handled by the compiler like you'd normally expect.
This change results in reduced restricted stack usage.

contracts/registrar/Registrar.sol
beardhero: Optimization. Further optimization must be decided on to change.

contracts/registrar/Registrar.sol
beardhero: Re-added unused variables.
Upgrade tests are out-of-scope.

contracts/oz/access/Ownable2Step.sol
beardhero: removed

contracts/opensea/OwnedRegistrant.sol
beardhero: removed

contracts/opensea/DefaultOperatorFilterer.sol
beardhero: removed

package.json
beardhero: > removed

package.json
beardhero: removed

hardhat.config.ts
beardhero: Copied mainnet provider below.

PR #62, opened 2022-11-28
Update OpenSea OperatorFilter

deep-quality-dev:
- Fixed unit tests for `Registrar`, `zNAResolver` for forked mainnet node
- Fixed revert error from `OperatorFilter` when transfer `domains`
- Fix prettier errors


contracts/registrar/Registrar.sol
```diff
@@ -125,7 +125,8 @@ contract Registrar is
     uint256 rootDomainId_,
     string calldata collectionName,
     string calldata collectionSymbol,
-    address zNSHub_
+    address zNSHub_,
+    address filterer_
```
JamesEarle: I had some technical challenges making this an alternate variable because of implications with testing and after fixing some size issues was able to get it back to an inherited model `Registrar is ... OperatorFilterer` .

It's your choice on leaving it this way or bringing back to the inheritance it was initially.

test/registrar.spec.ts
```diff
@@ -30,53 +34,36 @@ describe("Registrar", () => {
...
+      "Zer0 Name Service",
+      "ZNS",
+      hub.address,
+      operatorFilterer.address
+    );
+
+    await hub.addRegistrar(rootDomainId, registry.address);
   };
 
-  const blurAddress = "0x00000000000111AbE46ff893f3B2fdF1F759a8A8";
```
JamesEarle, line 58: Consider adding at least one of these values and it's matching test below again to show that sales revert when they are used

JamesEarle:
You also need to add an upgrade script to this PR for upgrading the default registrar as well as the various beacon registrars

contracts/registrar/Registrar.sol
```diff
@@ -174,7 +168,7 @@ contract Registrar is
       msg.sender == owner() || msg.sender == parentRegistrar,
       "ZR: Not authorized"
     );
-    require(!controllers[controller], "ZR: Controller is already added");
+    require(controllers[controller], "ZR: Controller is already added");
```
deep-quality-dev: This should be `require(!controllers[controller])`, right? @JamesEarle

contracts/registrar/Registrar.sol
deep-quality-dev: Yep I agree to use inherited model. I updated it

contracts/opensea/OperatorFilterer.sol
```diff
@@ -43,7 +43,21 @@ abstract contract OperatorFilterer {
...
+  }
+
+  function _unregister(address filter) internal {
+    require(_isRegistered(filter), "OF: Not registered");
+    OPERATOR_FILTER_REGISTRY.register(filter);
+  }
+
+  function _isRegistered(address filter) internal view returns (bool) {
+    return OPERATOR_FILTER_REGISTRY.isRegistered(filter);
+  }
```
beardhero: These functions are not needed as they can already be run on the OperatorFilterRegistry by the owner.

The owner passes the onlyAddressOrOwner(registrant) check, code link - https://github.com/ProjectOpenSea/operator-filter-registry/blob/adbb12ef37519e674cfc3482c9a43d7db7a1b31f/src/OperatorFilterRegistry.sol#L85

deep-quality-dev: Good point, it takes a bit of code size, I don't think we need some functions (i.e. register, unregister, isRegistered) of behalf of filter registry. What you think @JamesEarle

JamesEarle: I had them there to be able to call directly from the registrar not from the registry, but yes they caused file size issues and were out of scope, from what Joel said, so they've been removed

contracts/registrar/Registrar.sol
JamesEarle: yes I fixed it, my mistake not replacing it

contracts/opensea/OperatorFilterer.sol
```diff
@@ -20,70 +20,44 @@ contract OperatorFilterer {
   function initializeFilter(
     address subscriptionOrRegistrantToCopy,
     bool subscribe
-  ) public {
+  ) internal {
     // If an inheriting token contract is deployed to a network without the registry deployed, the modifier
     // will not revert, but the contract will need to be registered with the registry once it is deployed in
     // order for the modifier to filter addresses.
-    if (address(OPERATOR_FILTER_REGISTRY).code.length > 0) {
```
JamesEarle, line 27: Add a comment noting this check was removed

contracts/opensea/OperatorFilterer.sol
```diff
@@ -11,79 +11,62 @@ import {IOperatorFilterRegistry} from "./IOperatorFilterRegistry.sol";
...
-      if (
-        !OPERATOR_FILTER_REGISTRY.isOperatorAllowed(address(this), operator)
-      ) {
-        revert OperatorNotAllowed(operator);
-      }
+  function _onlyAllowedOperatorApproval(address operator) internal virtual {
+    if (!OPERATOR_FILTER_REGISTRY.isOperatorAllowed(address(this), operator)) {
+      revert OperatorNotAllowed(operator);
     }
-    _;
```
RyanCross, line 87: what is this about?

deep-quality-dev: It means that runs the main logic where the modifier is located in, only available in modifier.

PR #63, opened 2022-12-13
Update OpenSea OperatorFilter to Registrar

deep-quality-dev:
* Inherit from OperatorFilter abstract contract
* Reduce contract size
* Ensure all the test scripts works, including OperatorFilter test cases
* Update script
* Update npm packages


contracts/registrar/Registrar.sol
```diff
@@ -476,116 +531,92 @@ contract Registrar is
...
-    override
-    returns (uint256)
-  {
-    uint256 amount = records[id].royaltyAmount;
-    return amount;
+  function domainRoyaltyAmount(
+    uint256 id
+  ) public view override returns (uint256) {
+    return records[id].royaltyAmount;
+    // return domainManager._domainRoyaltyAmount(id);
```
JamesEarle: remove commented code

contracts/registrar/Registrar.sol
```diff
@@ -602,24 +633,47 @@ contract Registrar is
...
+    if (!controllers[msg.sender] && !zNSHub.isController(msg.sender)) {
+      revert NotController();
+    }
+  }
+
   function _setDomainMetadataUri(uint256 id, string memory uri) internal {
-    require(records[id].domainGroup == 0, "Must update via domain group");
+    if (records[id].domainGroup != 0) {
+      revert ShouldUpdateViaDomainGroup();
+    }
```
JamesEarle, line 660: This function should also check to make sure the token exists and that people aren't calling to set the same value for the URI

```solidity
if (tokenUri(id) == uri) { // the call to public function `tokenUri` will perform an `_exists` check
  revert SameMetadataUri(); // this error must be made
}
```

package.json
```diff
@@ -1,27 +1,29 @@
...
     "@typescript-eslint/eslint-plugin": "4.16.1",
     "@typescript-eslint/parser": "4.16.1",
+    "@zero-tech/zero-contracts": "0.0.7",
     "chai": "4.3.6",
     "dotenv": "8.2.0",
     "eslint": "7.21.0",
     "eslint-plugin-prettier": "3.3.1",
     "ethereum-waffle": "^3.4.0",
-    "ethers": "5.5.4",
+    "ethers": "^5.7.2",
```
JamesEarle: remove `^`

text.txt
```diff
@@ -0,0 +1,6 @@
+yarn run v1.22.17
```
JamesEarle: this file looks like it should be deleted

contracts/registrar/Registrar.sol
```diff
@@ -8,14 +8,37 @@ import {IRegistrar} from "../interfaces/IRegistrar.sol";
...
 contract Registrar is
   IRegistrar,
   OwnableUpgradeable,
-  ERC721PausableUpgradeable
+  ERC721PausableUpgradeable,
+  OperatorFilterer
 {
   using EnumerableMapUpgradeable for EnumerableMapUpgradeable.UintToAddressMap;
 
+  error InvalidFolderGroup();
```
JamesEarle, line 22: Like I mentioned in DMs this is a plus because it reduces gas and bytecode size for the contract but eliminates the reason strings when things fail. Like you mentioned we can get more information by simulating the tx in tenderly but when debugging with hardhat it remains an opaque error and can cause things to be unclear in our tests. We should review how to address this properly

contracts/registrar/Registrar.sol
```diff
@@ -140,6 +162,10 @@ contract Registrar is
 
     __ERC721Pausable_init();
     __ERC721_init(collectionName, collectionSymbol);
+    _initializeFilter(
+      address(0x3cc6CddA760b79bAfa08dF41ECFA224f810dCeB6),
```
JamesEarle, line 169: Might be nice to have a comment here linking to the opensea docs explaining this hardcoded address

contracts/registrar/Registrar.sol
```diff
@@ -447,20 +486,37 @@ contract Registrar is
   ) public {
     for (uint256 i = 0; i < tokenIds.length; ++i) {
       uint256 tokenId = tokenIds[i];
-      require(
-        _isApprovedOrOwner(_msgSender(), tokenId),
-        "ERC721: transfer caller is not owner nor approved"
-      );
+      if (!_isApprovedOrOwner(_msgSender(), tokenId)) {
```
JamesEarle, line 492: can just do `if (!_isApprovedOrOwner(_msgSender(), tokenIds[i]))` instead of assigning to a variable (inverse to my other comment before about not doing this when you're writing TypeScript, but this is Solidity)

test/domainPurchaser.spec.ts
```diff
@@ -31,8 +29,8 @@ describe("DomainPurchaser", () => {
   let creator: SignerWithAddress;
   let user1: SignerWithAddress;
 
-  const rootDomainHash = ethers.constants.HashZero;
-  const rootDomainId = BigNumber.from(0);
+  // const rootDomainHash = ethers.constants.HashZero;
```
JamesEarle: remove commented code

test/zNShub.spec.ts
```diff
@@ -83,14 +80,14 @@ describe("zNS Hub", () => {
...
-      console.log(creator);
+      // console.log(creator);
       const hubFactory = new ZNSHub__factory(creator);
       hub = await hubFactory.deploy();
     });
 
     it("can transfer ownership", () => {
       console.log("ZZZ");
-      console.log(hub);
+      // console.log(hub);
```
JamesEarle: remove console log

JamesEarle (commented):
Just a few cleanliness things

contracts/registrar/Registrar.sol
JamesEarle: Thanks for adding this in the other PR, I will resolve this comment.

contracts/registrar/Registrar.sol
deep-quality-dev: I don't think so, there are two uses of `tokenId` in this function.
instead of using storage variable, assigning to the memory variable and using takes a little gas fee.
Here's the example to show different gas fees.

```solidity
contract Assign {
    uint256 value;

    function assign(uint256 value_) external {
        value = value_;
    }

    function test() external {
        uint256 temp = value;
        require(temp > 10, "above 10");
        inter(temp);
    }

    function test2() external {
        require(value > 10, "above 10");
        inter(value);
    }

    function inter(uint256 temp) internal {
        temp++;
    }
}
```
`test` function takes `23537` gas, `test2` function takes `23602` gas.

JamesEarle: Tested myself and you're correct, so nevermind. Thanks for verifying

JamesEarle (approved):

PR #65, opened 2022-12-14
Catch custom error in test scripts

deep-quality-dev:


JamesEarle (approved):
Thanks for making sure we can see this on hardhat

PR #67, opened 2022-12-30
Create root domains for Degen

deep-quality-dev:
Create root domains: 0://degen, 0://dgen

This PR should be updated again with domain meta data.

### DIFF
--- a/scripts/degen/create-root-domains.ts
+++ b/scripts/degen/create-root-domains.ts
@@ -1,0 +1,82 @@
+import * as hre from "hardhat";
+import { ethers } from "hardhat";
+
+import { Registrar__factory } from "../../typechain";
+import { getLogger } from "../../utilities";
+import { zer0ProtocolAddresses } from "@zero-tech/zero-contracts";
+import { confirmContinue } from "../shared/helpers";
+import { domainConfigs } from "./config";
+
+const logger = getLogger("scripts::create-degen-root-domain");
+
+const main = async () => {
+  const [deployer] = await hre.ethers.getSigners();
+
+  logger.log(`Using deployer address ${deployer.address}`);
+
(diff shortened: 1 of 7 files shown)


scripts/degen/create-root-domains.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      const tx = await registrar.registerDomain(
+        parentId,
+        domainConfig.label,
+        domainConfig.minter,
+        domainConfig.metadataUri,
+        0,
+        false
+      );
+      await tx.wait();
```
JamesEarle: without providing a param, `tx.wait()` will loop indefinitely if you're using hardhat for any testing. Try something like

```typescript
const network = hre.network.name
const waitBlocks = network === "hardhat" ? 0 : 3;

...

await tx.wait(waitBlocks)
```

scripts/degen/create-root-domains.ts
```diff
@@ -0,0 +1,82 @@
...
+        },
+        {
+          Label: "Lock",
+          Info: false,
+        },
+      ]);
+
+      confirmContinue();
+
+      const tx = await registrar.registerDomain(
```
JamesEarle: After the root domain migration occurs all top-level domains will be beacon proxies instead and not regular domains within the default registrar, shown [here](https://www.notion.so/zerotech/Root-Domain-Migration-14b83742575740a2a3b7add1f24e20c0)

If we mint this domain as is, we will have to also include this domain in the migration scripts, but if we instead mint this domain as a beacon proxy through  `registerSubdomainContract()` we don't have to care about when it occurs relative to the migration, as it is already a beacon proxy

deep-quality-dev: OK, thanks for your information, I will update it with `registerSubdomainContract`

jtulloch (approved):

PR #71, opened 2023-02-09
NPM Package Setup (WIP)

Whytecrowe:
Requirements:
1. Create better folder structure
2. Create proper package for the NPM so it can be used in other repos
3. Set up tagging for both GitHub and NPM package versions
4. Add release flow to CircleCI to create a new tag on every `master` merge after all tests/builds pass
5. Test package in another repo to make sure we have everything needed
6. Update from `master` and fix conflicts before merge


package.json
```diff
@@ -1,4 +1,35 @@
 {
+  "name": "zns-contracts",
+  "version": "0.0.0-test.1",
```
Will-Hyde, line 3: Why use a different version?
Also, I think we normally follow this versioning scheme in our packages:
https://en.wikipedia.org/wiki/Software_versioning#Semantic_versioning

Will-Hyde:
What is the purpose of moving the archived scripts into the src directory? Do we plan on including them in the NPM package?

Whytecrowe:
> What is the purpose of moving the archived scripts into the src directory? Do we plan on including them in the NPM package?

@Will-Hyde no, we will delete them. I didn't know at the time what these were, so I kept them for the time being.

package.json
Whytecrowe: this is just for testing to make sure versions are aligned in all places. it will obviously change. this is not gonna be it :)
and yes, semver is good

PR #72, opened 2023-02-10
feat: isolate unit test, deploy Registrar, zNSHub on hardhat node

deep-quality-dev:
- Deprecate hardhat node configuration
- Deploy zNSHub, Registrar on hardhat node
-  Make sure run unit test successfully without forking node


test/helpers/deploy.ts
```diff
@@ -0,0 +1,231 @@
...
+    true
+  );
+  return true;
+};
+
+/**
+ * Deploy OperatorFilterRegistry at the address of "0x000000000000AAeB6D7670E522A718067333cd4E"
+ * https://github.com/ProjectOpenSea/operator-filter-registry
+ */
+const deployOperatorFilterRegistry = async (): Promise<void> => {
```
RyanCross: Some more context as to why this is needed for testing would be a good add here.  What problem does this solve? What issue does this get around?

test/helpers/deploy.ts
```diff
@@ -0,0 +1,231 @@
...
+    "0x4c8d290a1b368ac4728d83a9e8321fc3af2b39b1"
+  );
+  const user2 = await impersonateAddressWithBalance(
+    "0x2f7cf35d65ffcb50c919b24a730e6f8f4aa76f46"
+  );
+  const user3 = await impersonateAddressWithBalance(
+    "0x9aac739c133074db445183a95149880a2156541a"
+  );
+
+  await gotoStep1(user1);
```
JamesEarle: Please don't name functions like this. It's impossible to understand what this function does based on the current name.

test/helpers/deploy.ts
RyanCross: Also, impressed you figured this out, so props 👍 .  How long did it take to get this working? Do you foresee this type of testing limitation coming up in other repos?

test/helpers/deploy.ts
```diff
@@ -0,0 +1,231 @@
...
...928301207fff000000000000000000000000000000000000000000000000000000000000008383015260609990991b7fffffffffffffffffffffffffffffffffffffffff000000000000000000000000166021820152603581019790975260558088019890985282518088039098018852607590960182525085519585019590952073ffffffffffffffffffffffffffffffffffffffff81166000908152948590529490932054939450505060ff909116159050610697575060005b9392505050565b604080517fff000000000000000000000000000000000000000000000000000000000000006020808301919091523060601b6021830152603582018590526055808301859052835180840390910181526075909201835281519181019190912073ffffffffffffffffffffffffffffffffffffffff81166000908152918290529190205460ff161561072e575060005b9291505056fe496e76616c696420636f6e7472616374206372656174696f6e202d20636f6e74726163742068617320616c7265616479206265656e206465706c6f7965642e496e76616c69642073616c74202d206669727374203230206279746573206f66207468652073616c74206d757374206d617463682063616c6c696e6720616464726573732e4661696c656420746f206465706c6f7920636f6e7472616374207573696e672070726f76696465642073616c7420616e6420696e697469616c697a6174696f6e20636f64652ea265627a7a723058202bdc55310d97c4088f18acf04253db593f0914059f0c781a9df3624dcef0d1cf64736f6c634300050a0032",
```
RyanCross: What exactly am I looking at here 😄

contracts/mocks/OwnedRegistrant.sol
```diff
@@ -0,0 +1,78 @@
...
+ * there is an account (an owner) that can be granted exclusive access to
+ * specific functions.
+ *
+ * By default, the owner account will be the one that deploys the contract. This
+ * can later be changed with {transferOwnership} and {acceptOwnership}.
+ *
+ * This module is used through inheritance. It will make available all functions
+ * from parent (Ownable).
+ */
+abstract contract Ownable2Step is Ownable {
```
RyanCross: More docs pls. Why is this necessary?

The context of why this is here will be invaluable in 6 months when we're looking at this going 🤔

test/helpers/deploy.ts
deep-quality-dev: Our `Registrar` is interacting with `OperatorFilterRegistry` which was deployed at specific address, as well as in unit test.
So to deploy `Registrar`, it requires to deploy the OpenSea contracts first.
I tracked the contract creation transactions on Etherscan, BscScan, AvaScan, and tried the all the transactions. :) because that contract was deployed at the same address across the chain.
To deploy contract at specific address, we can use `CREATE2` opcode, it have to pass `sender address`, `bytecode` and `salt`.
So I impersonate the account like in etherscan and send same byte data to the node to deploy contracts.
Now it works, ya 👍

contracts/mocks/OwnedRegistrant.sol
deep-quality-dev: I added `Ownable2Step` contract into `mock` folder, this contract is used to transfer ownership which was created by `CREATE2` opcode.
`CREATE2` opcode is only callable by contract, so contract can deploy contract, and the ownership of new contract will be the deployer contract.
Once deployed, by calling `acceptOwnership` of `Ownable2Step`, the ownership will be transfered back to the `msg.sender`

Whytecrowe: let's add work "Mock" at the end of the name for every Mock contract. this will help us avoid confusion and errors later when integration tests and such are added. we won't see folder structure from everywhere to determine that it's a mock

hardhat.config.ts
```diff
@@ -89,18 +89,18 @@ const config: HardhatUserConfig = {
...
-      //     privateKey: `0x${process.env.TESTNET_PRIVATE_KEY}`,
-      //     balance: "10000000000000000000000",
-      //   },
-      // ],
-      forking: {
-        url: `https://goerli.infura.io/v3/0e6434f252a949719227b5d68caa2657`,
-      },
-      allowUnlimitedContractSize: true
-    },
+    // hardhat: {
```
Whytecrowe: is it possible to keep this flow, but make it run with a specific script that is not used in CI? so that we can still run this locally or on a server?
this might require setting up all the deploy helpers in a configurable way depending on the script. so if that's proven to be much work, let me know and mb we can move it to another task.

test/helpers/deploy.ts
Whytecrowe: dark magic :)

Whytecrowe: @deep-quality-dev do we even need this now if we are mocking the OS contract?

test/helpers/deploy.ts
Whytecrowe: This is just more OS problems...
Good job figuring this out! But this seems too hacky and not very sustainable. :)
Hopefully, the new solution you just talked about works.

test/helpers/deploy.ts
```diff
@@ -0,0 +1,231 @@
...
+  OwnedRegistrant__factory,
+  Registrar,
+  Registrar__factory,
+  UpgradeableBeacon__factory,
+  ZNSHub,
+  ZNSHub__factory,
+} from "../../typechain";
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
+
+export const impersonateAddressWithBalance = async (
```
JamesEarle: This doesn't look like it's used by anything beyond this file, so don't export it

test/registrar.spec.ts
```diff
@@ -76,7 +47,9 @@ describe("Registrar", () => {
 
   describe("root domain", () => {
     before(async () => {
-      await deployRegistry(creator);
+      const { registrar, zNSHub } = await deployZNS(creator);
+      registry = registrar;
```
Whytecrowe: let's call them the same name. registry assumes a different logic then registrar. so it's a bit confusing.
also, we can destructure and assign together as:
```javascript
({ registrar: registry, zNSHub: hub } = await deployZNS(creator));
```

or just:
```javascript
({ registrar, zNSHub } = await deployZNS(creator));
```
if our vars in test are the same name.

test/helpers/deploy.ts
deep-quality-dev: I found another solution to deploy contract at specific address by using `hardhat_setCode` command, already pushed.
Please check it again?

test/helpers/deploy.ts
deep-quality-dev: Yes, don't use it, we can deploy contract at specific address by using `hardhat_setCode`

contracts/mocks/OwnedRegistrant.sol
deep-quality-dev: Yes for sure, renamed

test/helpers/deploy.ts
```diff
@@ -0,0 +1,110 @@
...
+const deployOperatorFilterRegistry = async (deployer: SignerWithAddress): Promise<void> => {
+  const deployed = await isOperatorFilterRegistryDeployed();
+  if (deployed) {
+    return;
+  }
+
+  // Deploy OperatorFilterRegistry and get bytecode from it
+  const OperatorFilterRegistryFactory = new OperatorFilterRegistry__factory(deployer);
+  const operatorFilterRegistryInst = await OperatorFilterRegistryFactory.deploy();
+  const codeOperator = await ethers.provider.send("eth_getCode", [operatorFilterRegistryInst.address]);
```
Whytecrowe: do we have to deploy here to set the bytecode later? the one from the compiler does not work?

deep-quality-dev: Good point, it works, I will commit soon

deep-quality-dev: Good point, we have two solution, one is copying deployed bytecode or use `deployedBytecode` in artifacts.

test/helpers/deploy.ts
```diff
@@ -0,0 +1,123 @@
...
+  // Since `hardhat_setCode` does not set or copy storage, it requires to
+  // set storage at 0x0 to set owner of `OwnedRegistrant` contract.
+  // Do the same thing of `OwnedRegistrant` constructor
+
+  // Set `deployer` account as owner of `OwnedRegistrant` contract
+  const ownerSlot = "0x0";
+  const value = ethers.utils.hexlify(
+    ethers.utils.zeroPad(deployer.address, 32)
+  );
+  await ethers.provider.send("hardhat_setStorageAt", [
```
JamesEarle: nice!

test/zNShub.spec.ts
```diff
@@ -78,61 +26,61 @@ describe("zNS Hub", () => {
...
 
     it("can transfer ownership", () => {
-      //await hub["transferOwnership"](user1.address);
-      //expect(await hub.registrarBeacon()).to.eq(user1.address);
+      //await zNSHub["transferOwnership"](user1.address);
+      //expect(await zNSHub.registrarBeacon()).to.eq(user1.address);
     });
   });
 
   // describe("Controllers", () => {
```
JamesEarle: uncomment this and make sure these tests pass too

JamesEarle (approved):

RyanCross (approved):
Looking good. Thanks for the context

